PR #15, opened 2025-05-28
Initial code for wallet

stefanbitcr:
**Notes**
* Majority of code is placeholder
* Overly specialized for Wilcat credit mint
* Code refactor is likely when adding a indxb as persistent storage
* Send and receiving of credit tokens are possible, with them being swapped for new ones using Wildcat
* Lack of tests


Cargo.toml
```diff
@@ -0,0 +1,43 @@
+[workspace]
+resolver = "2"
```
zupzup, line 2: Is there a reason not to use resolver 3? (Rust 2024)

crates/bcr-wallet-core/scripts/test_mac.sh
```diff
@@ -0,0 +1,5 @@
+# Clang/LLVM issues on MAC
```
zupzup, line 1: Protom added some build and check commands for our wasm build to our [justfile](https://github.com/BitcreditProtocol/Bitcredit-Core/blob/master/justfile recently - maybe they can be useful here as well

Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+members = ["crates/bcr-wallet-core"]
+
+[workspace.dependencies]
+bcr-ebill-core = { git = "https://github.com/BitcreditProtocol/Bitcredit-Core.git", tag = "v0.3.11" }
+bcr-ebill-api = { git = "https://github.com/BitcreditProtocol/Bitcredit-Core.git", tag = "v0.3.11" }
+bcr-wdc-webapi = { git = "https://github.com/BitcreditProtocol/Wildcat" }
+anyhow = { version = "1" }
+async-trait = { version = "0.1" }
+axum = { version = "0.8" }
+axum-test = { version = "17" }
```
zupzup, line 13: axum? :thinking: 

I guess this is just a collection of crates we use generally to start with and select out later, right?

crates/bcr-wallet-core/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "bcr-wallet-core"
+version = "0.1.0"
+edition = "2024"
```
zupzup, line 4: +license=MIT (also in workspace Cargo.toml)

or directly use 

```
version.workspace = true
edition.workspace = true
license.workspace = true
```

and put all int ehr workspace

crates/bcr-wallet-core/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+serde.workspace = true
+async-trait.workspace = true
+log = "0.4.27"
+reqwest = { workspace = true, features = ["json"] }
+anyhow.workspace = true
+rand.workspace = true
+tracing.workspace = true
+tracing-subscriber.workspace = true
+tracing-wasm = "0.2.1"
+getrandom = { version = "0.2.16", features = ["js"] }
```
zupzup, line 24: I guess you had a build error and added this because of it? I'm just asking because that's quite an old version

crates/bcr-wallet-core/src/wallet/builder.rs
```diff
@@ -0,0 +1,102 @@
+// ----- standard library imports
+use std::marker::PhantomData;
+// ----- extra library imports
+// ----- local modules
+use super::connector::{Connector, MintConnector};
+use crate::db::WalletDatabase;
+use crate::wallet::{CreditWallet, DebitWallet, Wallet, WalletType};
+// ----- end imports
+
+#[derive(Debug)]
```
zupzup, line 10: hm I think this only applies to `Unconfigured`, right? :thinking:

crates/bcr-wallet-core/src/wallet/connector.rs
```diff
@@ -0,0 +1,107 @@
...
+impl MintConnector for Connector<CreditWallet> {
+    async fn list_keysets(&self) -> cdk02::KeysetResponse {
+        let url = self.url("v1/keysets");
+        self.client.get(url).await.unwrap()
+    }
+    async fn swap(&self, req: cashu::SwapRequest) -> Result<cashu::SwapResponse> {
+        let url = self.url("v1/swap");
+        self.client.post(url, &req).await
+    }
+    /// GET v1/keys/{kid}
```
zupzup: This comment seems a bit superfluous to me, since the url is directly under it.

if the methods are more complex, it def. makes sense, but then I would add it to all of them for consistency :+1:

crates/bcr-wallet-core/src/wallet/utils.rs
```diff
@@ -0,0 +1,73 @@
...
+                new_subset.push(proof.clone());
+                dp.insert(new_sum, new_subset);
+            }
+        }
+    }
+
+    dp.get(&send_amount).cloned()
+}
+
+pub fn get_amounts(mut targ: u64) -> Vec<u64> {
```
zupzup, line 32: while I very much appreciate using bitwise operators to split out the powers of two, I feel like a few comments on what this method does and why/how wouldn't hurt, since this stuff isn't used a lot in most Rust code (e.g. for external contributors)

crates/bcr-wallet-core/src/wallet/api.rs
```diff
@@ -0,0 +1,156 @@
...
+
+            let mut proofs = proofs;
+            proofs.retain(|p| !selected_cs.contains(&p.c));
+            self.db.set_proofs(proofs).await;
+
+            return token.to_v3_string();
+        } else {
+            warn!("Could not select subset of proofs to send");
+        }
+        "".into()
```
zupzup, line 154: Should this return an error in this case? empty string seems strange :thinking:

zupzup (approved):
Very nice :muscle:

crates/bcr-wallet-core/src/wallet/connector.rs
stefanbitcr: leftovers

crates/bcr-wallet-core/src/wallet/utils.rs
codingpeanut157, line 33: There already are APIs for such operation in cashu in `chashu::amount`
`cashu::Amount::split()` should do exactly what this algorithm does

crates/bcr-wallet-core/src/wallet/utils.rs
```diff
@@ -0,0 +1,73 @@
...
+
+pub fn generate_blind(
+    kid: cashu::Id,
+    amount: cashu::Amount,
+) -> (
+    cashu::BlindedMessage,
+    cashu::secret::Secret,
+    cashu::SecretKey,
+) {
+    let secret = cashu::secret::Secret::new(rand::random::<u64>().to_string());
```
codingpeanut157, line 73: we need to follow [NUT13](https://github.com/cashubtc/nuts/blob/main/13.md) to make the wallet recoverable

codingpeanut157 (approved):
A few changes required, I am not going to hold back this PR though.
Up to you to either fix as per comments or merge this and create issues

crates/bcr-wallet-core/src/wallet/utils.rs
stefanbitcr: I'll open a separate issue to track this

stefanbitcr:
> @stefanbitcr did you have a look at the comments made by copilot, I think some are interesting Should we leave a sticky note somewhere to remember this set of suggestions for future us?

@codingpeanut157 Yes I did (if you mean [github-advanced-security](https://github.com/BitcreditProtocol/Wallet-Core/security/code-scanning) ) , regarding the Guards and Locks I'm not sure if I will keep them or not, they are placeholder, (some) of the others I addressed.

PR #19, opened 2025-05-30
Add Rexie for persistent storage

stefanbitcr:
This update adds persistent storage for proofs

* Initial support for multiple wallets
* Adds rexie (IndexedDB) for storing proofs
* Proofs remain stored, and get marked as spent when swapped or sent, so that they can be later swept and potentially reclaimed.

**Issues**
* Rexie DB handle is included in the Wallet Database which is !Send, however the WASM code is single threaded.
* Wallets are created up front, each wallet uses its own store
* Filters for unspent proofs inefficiently 
* Error  handling, Wallet configurations such as minturl, keyset associations with mint


mtbitcr (approved):

crates/bcr-wallet-core/src/db/types.rs
```diff
@@ -1,13 +1,34 @@
...
+    pub(crate) status: ProofStatus,
+    pub(crate) id: PublicKey, // This might change
+}
+
 pub trait WalletDatabase {
-    async fn get_proofs(&self) -> Vec<Proof>;
-    async fn set_proofs(&mut self, proofs: Vec<Proof>);
-    async fn add_proof(&mut self, proof: Proof);
+    async fn get_active_proofs(&self) -> Result<Vec<Proof>, DatabaseError>;
+    async fn inactivate_proof(&self, proof: Proof) -> Result<(), DatabaseError>;
```
zupzup, line 32: inactivate sounds a bit strange to me :sweat_smile: maybe `deactivate`? Or is there a specific reason this is called `inactivate`?

crates/bcr-wallet-core/src/db/rexie/mod.rs
```diff
@@ -0,0 +1,107 @@
+// ----- standard library im
+use std::rc::Rc;
+// ----- extra library imports
+// use anyhow::Result;
```
zupzup, line 4: can be removed

zupzup (approved):
Nice, LGTM :+1:

crates/bcr-wallet-core/src/db/types.rs
zupzup: hm, or just really specifically "set_to_spent", because that's what actually happens, right?

stefanbitcr: > hm, or just really specifically "set_to_spent", because that's what actually happens, right?

I was hesitant to say its spent because it might still get marked even though we send it to someone, so that the wallet won't use it immediately in another TX.

A better solution is to have another state, maybe pending for our guesses, and spent if we actually know.

But I'll add that once we have more functionality.

Issue #20, opened 2025-06-02
Implement NUT07 States


stefanbitcr:
Check whether proofs are [UNSPENT, PENDING, SPENT](https://github.com/cashubtc/nuts/blob/main/07.md)

Periodically check whether PENDING proofs have been spent and mark them accordingly.

PR #21, opened 2025-06-04
Adds deterministic secrets for blinded message generation and support for recovery

stefanbitcr:
Implements NUT 13 for secret generation using Cashu CDK, and NUT09 for recovery.


crates/bcr-wallet-core/src/wallet/api.rs
```diff
@@ -105,6 +53,118 @@ where
...
+            let mut fruitless_attempts = 0;
+            let mut key_counter = 0;
+
+            while fruitless_attempts < 3 {
+                let premint_secrets = cashu::PreMintSecrets::restore_batch(
+                    kid,
+                    self.xpriv,
+                    key_counter,
+                    key_counter + 100,
+                )?;
```
zupzup, line 100: should this fail the whole function, or should it also continue on error?

crates/bcr-wallet-core/src/wallet/api.rs
```diff
@@ -105,6 +53,118 @@ where
...
+                let proofs = cashu::dhke::construct_proofs(
+                    response.signatures,
+                    premint_secrets.iter().map(|p| p.r.clone()).collect(),
+                    premint_secrets.iter().map(|p| p.secret.clone()).collect(),
+                    &keys,
+                )?;
+
+                self.db.increase_count(kid, proofs.len() as u32).await?;
+
+                let ys: Vec<cashu::PublicKey> = proofs.iter().map(|p| p.y().unwrap()).collect();
```
zupzup, line 140: Is this unwrap temporary / are these y's guaranteed to be there?

crates/bcr-wallet-core/src/wallet/api.rs
```diff
@@ -105,6 +53,118 @@ where
...
+    }
+
+    pub async fn restore(&self) -> anyhow::Result<()> {
+        let keysets = self.connector.list_keysets().await?;
+
+        let keyset_ids: Vec<cashu::Id> = keysets.keysets.iter().map(|ks| ks.id).collect();
+
+        let mut restored_value = cashu::Amount::ZERO;
+
+        for kid in keyset_ids {
```
zupzup, line 82: a comment could be helpful here, which describes the used algorithm (fruitless attempts / increasing key_counter)

stefanbitcr: Good point. In Nut13 of Cashu they describe restoring proofs in batches of 100, and if there are 3 batches where nothing is restored, we stop.

I'll add a similar comment.

mtbitcr (approved):

crates/bcr-wallet-core/src/wallet/api.rs
stefanbitcr: The ys are basically a hashed version of the secrets, I can move it out so we fail if there is something wrong (which isn't something that should happen).

crates/bcr-wallet-core/src/wallet/api.rs
stefanbitcr: CDK inspired, even looking at the source, its not immediately clear what triggers secret generation to fail under standard conditions. Its fairly sensible to stop if there is something wrong, which might relate to our private key.

crates/bcr-wallet-core/src/db/rexie/wallet.rs
```diff
@@ -0,0 +1,150 @@
...
+        )?;
+        let store = tx.store(&self.store_name.clone())?;
+
+        let key = proof
+            .y()
+            .map_err(|e| DatabaseError::CdkError(e.to_string()))?;
+        let key = to_js(&key)?;
+        if let Ok(Some(wp)) = store.get(key).await {
+            let mut wp: WalletProof = from_js(wp)?;
+            wp.status = ProofStatus::Spent;
```
codingpeanut157, line 79: If possible
I would use [cashu::nut07::State](https://docs.rs/cashu/latest/cashu/nuts/nut07/enum.State.html) and change this to the proper state, probably `PendingSpent` in this case.
We might also want to change the function name to something more specific for this use case, or maybe add the state as a parameter.

codingpeanut157 (commented):
Looks good
left a few comments in code :+1:

crates/bcr-wallet-core/src/db/rexie/wallet.rs
```diff
@@ -0,0 +1,150 @@
...
+use serde::de::DeserializeOwned;
+use serde_wasm_bindgen::{from_value, to_value};
+use wasm_bindgen::JsValue;
+// ----- local modules
+use crate::db::types::{DatabaseError, ProofStatus, WalletProof};
+use crate::db::{KeysetDatabase, WalletDatabase};
+// ----- end imports
+
+pub struct RexieWalletDatabase {
+    db: Rc<Rexie>,
```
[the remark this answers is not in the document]
stefanbitcr: Yes, we can unsafe implement Send and Sync if we need it

crates/bcr-wallet-core/src/app.rs
```diff
@@ -31,12 +31,13 @@ thread_local! {
...
 
 pub async fn initialize() {
-    let manager = db::Manager::new().await.unwrap();
+    let manager = db::rexie::Manager::new().await.unwrap();
 
     let mut wallets = Vec::new();
     for i in 0..10 {
         let rexie_wallet = RexieWalletDatabase::new(format!("wallet_{}", i), manager.get_db());
-        let mint_url = MintUrl::from_str("https://wildcat-dev-docker.minibill.tech").unwrap();
+        // let mint_url = MintUrl::from_str("https://wildcat-dev-docker.minibill.tech").unwrap();
```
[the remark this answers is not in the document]
stefanbitcr: Probably not.

It's a symptom from manually changing it.

I'll remove this next when the user can specify the wallets, mint url, mnemonic and so on.

crates/bcr-wallet-core/src/db/rexie/wallet.rs
stefanbitcr: There is a ticket for this 

https://github.com/BitcreditProtocol/Wallet-Core/issues/20

codingpeanut157 (approved):

crates/bcr-wallet-core/src/db/rexie/wallet.rs
codingpeanut157: Yeah, they might need an extra state or so for their internal operation.
The key point though is that tokens are not spent (and thus deletable from the DB) right after a swap or a payment, but they rather go into a `Pending` state that must be checked later against the `tokenstatecheck` endpoint.

PR #23, opened 2025-06-09
Add initial redemption support

stefanbitcr:
### Changes

* Allows users to specify wallets
* Adds initial debit support
* Allows credit wallet to sweep inactive keyset

### DIFF
--- a/crates/bcr-wallet-core/src/db/rexie/metadata.rs
+++ b/crates/bcr-wallet-core/src/db/rexie/metadata.rs
@@ -1,0 +1,127 @@
+// ----- standard library im
+use std::rc::Rc;
+// ----- extra library imports
+use cashu::MintUrl;
+use rexie::Rexie;
+use rexie::TransactionMode;
+// ----- local modules
+use super::utils;
+use crate::db::rexie::utils::from_js;
+use crate::db::types::DatabaseError;
+use crate::db::{Metadata, WalletMetadata};
+// ----- end imports
+
+pub struct RexieMetadata {
+    db: Rc<Rexie>,
+}
+
+impl RexieMetadata {
+    pub fn new(db: Rc<Rexie>) -> Self {
+        Self { db }
+    }
+
+    async fn len(&self) -> Result<usize, DatabaseError> {
+        let tx = self.db.transaction(
+            std::slice::from_ref(&super::constants::WALLET_METADATA),
+            TransactionMode::ReadOnly,
+        )?;
+
+        let store = tx.store(super::constants::WALLET_METADATA)?;
+        let keys = store.get_all_keys(None, None).await?;
+
+        tx.done().await?;
+
+        Ok(keys.len())
+    }
+}
+
+impl Metadata for RexieMetadata {
+    async fn get_wallets(&self) -> Result<Vec<WalletMetadata>, DatabaseError> {
+        let tx = self.db.transaction(
+            std::slice::from_ref(&super::constants::WALLET_METADATA),
+            TransactionMode::ReadOnly,
+        )?;
+
+        let store = tx.store(super::constants::WALLET_METADATA)?;
+
+        let wallets = store
+            .get_all(None, None)
+            .await?
+            .into_iter()
+            .map(utils::from_js)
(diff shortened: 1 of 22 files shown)


zupzup (approved):
LGTM :+1:

crates/bcr-wallet-core/src/db/rexie/metadata.rs
```diff
@@ -0,0 +1,142 @@
...
+    db: Rc<Rexie>,
+}
+
+impl RexieMetadata {
+    pub fn new(db: Rc<Rexie>) -> Self {
+        Self { db }
+    }
+
+    /// Finds the first unoccupied natural number to use as id
+    async fn get_empty_id(&self) -> Result<usize, DatabaseError> {
```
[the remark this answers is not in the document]
stefanbitcr: Good point !

mtbitcr (approved):

crates/bcr-wallet-core/src/wallet/credit.rs
```diff
@@ -1,85 +1,119 @@
...
-            warn!("No proofs provided");
-            return Err(anyhow::anyhow!("No proofs provided"));
-        };
-        let mut total_proofs = 0u64;
-        let mut total_amounts = 0u64;
+impl<DB> Wallet<CreditWallet, DB>
+where
+    DB: WalletDatabase + KeysetDatabase,
+{
+    pub async fn redeem_first_inactive(&self) -> anyhow::Result<String> {
```
codingpeanut157, line 19: :thinking: 
I don't get this function

Why do you want to redeem only one keyset at a time?
`redeem` endpoint accepts inputs with different keysets as long as they all are inactive.
Also, by redeeming multiple inactive keysets at once we could save swap operations and storage space as we need store less proofs

crates/bcr-wallet-core/src/wallet/api.rs
```diff
@@ -205,12 +208,82 @@ where
...
+    ) -> anyhow::Result<Vec<Proof>> {
+        let wdc = &self.connector;
+
+        if proofs.is_empty() {
+            warn!("No proofs provided");
+            return Err(anyhow::anyhow!("No proofs provided"));
+        }
+
+        let mut total_proofs = 0u64;
+        let mut total_amounts = 0u64;
```
codingpeanut157, line 234: I'd use `cashu::Amount` and `cashu::Amount::checked_add`

crates/bcr-wallet-core/src/wallet/api.rs
```diff
@@ -205,12 +208,82 @@ where
...
+        let rs = premint_secrets
+            .iter()
+            .map(|b| b.r.clone())
+            .collect::<Vec<_>>();
+
+        let proofs = cashu::dhke::construct_proofs(response.signatures, rs, secrets, &keys.keys)?;
+
+        let _ = self
+            .db
+            .increase_count(keyset_id, proofs.len() as u32)
```
codingpeanut157, line 284: I think it's safer to increase the counter before making the request.
If we do it before, and then something goes wrong, the worst that can happen is a gap in between indexes in the derivation path.
If we do it later, as of now, we get an error from the mint saying that it can't swap to tokens that it has already saw in the past.

stefanbitcr: Good point.

crates/bcr-wallet-core/src/wallet/credit.rs
```diff
@@ -1,85 +1,119 @@
...
+        // Credit tokens need to handle the edge case where the token we receive
+        // contains different keysets - we can only swap to the same keyset
 
-        let _ = self
-            .db
-            .increase_count(keyset_id, proofs.len() as u32)
-            .await?;
+        // Group proofs by keyset_id
+        let mut keyset_proofs = HashMap::new();
+        for p in &proofs {
```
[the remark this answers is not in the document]
stefanbitcr: Nice catch !

codingpeanut157 (approved):

stefanbitcr:
Thank you both @zupzup @codingpeanut157 !

PR #45, opened 2025-07-21
cleanDB + fix reclaim funds

codingpeanut157:


crates/bcr-wallet-core/src/pocket.rs
```diff
@@ -1164,4 +1191,67 @@ mod tests {
         let response = crpocket.prepare_send(amount, &k_infos).await;
         assert!(matches!(response, Err(Error::InsufficientFunds)));
     }
+
+    #[tokio::test]
+    async fn swap_proof_to_target() {
+        let (_, keyset) = keys_test::generate_keyset();
+        // 16 --> 13 ==> ( 8 + 4 + 1 ) + 2 +1
```
zupzup, line 1198: `+1` => `+ 1` for consistency

crates/bcr-wallet-core/src/app.rs
```diff
@@ -89,7 +92,8 @@ pub async fn add_wallet(name: String, mint_url: String, mnemonic: String) -> Res
     }
 
     // building database and object_stores
-    let rexie_db_name = format!("bitcredit_wallet_{}", mint_id.as_byte_array().as_hex());
+    let db_id = sha256::Hash::from_engine(hasher);
```
zupzup, line 95: This would still lead to the same hash, if the same mnemonic is used with the same e.g. name (or mint url / pub key) twice, right?

Is there an error, if this happens?

codingpeanut157: I guess no
if I use the same mnemonic, and I point to the same mint_url, I would expect the wallet to pick any pre-existing local DB.

zupzup: OK, so this wouldn't "override" the existing Wallet, but rather simply return the already existing one?
If so, this would be good to somehow encode/document, I guess, since `add_wallet` doesn't really communicate this.

From a UX-perspective, I guess it would be good to catch this somehow (not that important right now)

codingpeanut157: :thinking: 
rexie does not return any information regarding whether or not the db was newly created or if it was opened from an existing one.

I guess at some point this function is going to be refactored and adhere more to what the name expresses by deleting any pre-existing db with the same ID.

stefanbitcr (approved):
Neat ! 👍

PR #51, opened 2025-07-29
[bcr-wallet-core] add Purse concept

codingpeanut157:
In our architecture, 
a Pocket deals with a specific currency unit in a mint,
a Wallet deals with credit and debit pockets within a specific mint
while a `Purse` is a container of different wallets.

At the moment, the Purse is just a container of wallet configurations, used to persist wallet and mint information in local DB.
AppState is still in control of the wallets and their operations.
In future Purse will take charge of wallets and deal with intermint swaps


mtbitcr (approved):

codingpeanut157:
> Are credit wallets optional to support debit only Cashu mints ?

yes

PR #71, opened 2025-08-27
[bcr-wallet-core] refactor of the payment APIs

codingpeanut157:
- remove prepare_send and sed functions
- add prepare_payment and pay that allow nut18 and bolt11 payments

- add prepare_payment_request and check_received_payment for nut18 PaymentRequest

### DIFF
--- a/crates/bcr-wallet-core/src/purse.rs
+++ b/crates/bcr-wallet-core/src/purse.rs
@@ -1,15 +1,29 @@
 // ----- standard library imports
-use std::sync::{Arc, Mutex};
+use std::{
+    collections::HashMap,
+    sync::{Arc, Mutex},
+};
 // ----- extra library imports
 use async_trait::async_trait;
+use cashu::{Amount, CurrencyUnit, MintUrl, nut00 as cdk00, nut18 as cdk18};
+use cdk::wallet::types::TransactionId;
+use nostr_sdk::nips::{
+    nip19::{Nip19Profile, ToBech32},
+    nip59::UnwrappedGift,
+};
+use uuid::Uuid;
 // ----- local imports
-use crate::{error::Result, types::WalletConfig};
+use crate::{
+    error::{Error, Result},
+    sync,
+    types::{PaymentSummary, WalletConfig},
+};
 
 // ----- end imports
 
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
-pub trait PurseRepository {
+pub trait PurseRepository: sync::SendSync {
     async fn store(&self, wallet: WalletConfig) -> Result<()>;
     async fn load(&self, wallet_id: &str) -> Result<WalletConfig>;
     #[allow(dead_code)]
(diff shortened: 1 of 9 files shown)


crates/bcr-wallet-core/src/purse.rs
```diff
@@ -80,4 +137,142 @@ where
...
+        return Ok(false);
+    };
+    let cdk18::PaymentRequestPayload {
+        id,
+        mint,
+        proofs,
+        unit,
+        memo,
+    } = payload;
+    if id.unwrap_or_default() != payment_id {
```
zupzup, line 255: There is the potential case where `payment_id` is "" and the `id` is empty, in which case it wouldn't early exit here.
Would this be an issue, or is this scenario caught by other validation elsewhere?

crates/bcr-wallet-core/src/purse.rs
```diff
@@ -80,4 +137,142 @@ where
...
+        return Ok(false);
+    }
+    let wlt = {
+        let locked = wlts.lock().unwrap();
+        let found = locked.iter().find(|w| w.mint_url() == mint).cloned();
+        if found.is_none() {
+            return Ok(false);
+        }
+        found.unwrap()
```
zupzup, line 264: maybe `expect("checked above")` or somethign for clarity

crates/bcr-wallet-core/src/wallet.rs
```diff
@@ -118,30 +106,34 @@ pub trait TransactionRepository {
...
     pub tx_repo: TxRepo,
-    pub url: cashu::MintUrl,
+    pub mint_url: cashu::MintUrl,
     pub debit: DebtPck,
     pub credit: Box<dyn CreditPocket>,
     pub name: String,
     pub id: String,
     pub mnemonic: bip39::Mnemonic,
-    pub current_send: Mutex<Option<SendReference>>,
+    // pub current_send: Mutex<Option<SendReference>>,
```
zupzup, line 136: leftover?

crates/bcr-wallet-core/src/wallet.rs
```diff
@@ -160,26 +152,58 @@ where
...
         };
-        let summary = SendSummary {
-            request_id: reference.request_id,
-            amount,
-            unit: pocket.unit(),
-            send_fees: pocket_summary.send_fees,
-            swap_fees: pocket_summary.swap_fees,
+        let empty = vec![];
+        let transports = req.transports.as_ref().unwrap_or(&empty);
+        let (nostrs, https): (Vec<_>, Vec<_>) = transports
```
zupzup, line 183: isn't it `nostr` and `https`? or is the `s` for plurality of httpS and nostrS?
I guess it's for plurality, but I was a bit confused because `https` has a distinct meaning 😅 maybe `nostr_tsports`/`http_tsports` or smth?

Not too important, though

zupzup (approved):
Looks good to me 👍, but it would definitely be good if @tompro would check the Nostr parts for pitfalls

crates/bcr-wallet-core/src/purse.rs
codingpeanut157: `payment_id` is created by wallet-core itself at the `prepare_pay_request` stage, so it should never be empty.
If, on the other hand, we receive a random payment with `id` empty, we act kindly and we simply discard it and not cash it in.

crates/bcr-wallet-core/src/purse.rs
```diff
@@ -80,4 +137,160 @@ where
...
+            let wlts = self.wallets.lock().unwrap();
+            let mut mints = Vec::with_capacity(wlts.len());
+            for wlt in wlts.iter() {
+                mints.push(wlt.mint_url());
+            }
+            mints
+        };
+        let nostr_transport = cdk18::Transport {
+            _type: cdk18::TransportType::Nostr,
+            target: self.myself.to_bech32()?,
```
tompro, line 187: Does this mean we send the payment to ourself?

crates/bcr-wallet-core/src/purse.rs
```diff
@@ -80,4 +137,160 @@ where
...
+        let cl = Arc::clone(&self.nostr_cl);
+        let tx_id: Arc<Mutex<Option<TransactionId>>> = Arc::new(Mutex::new(None));
+        let notif_handler = self.nostr_cl.handle_notifications(|notif| async {
+            let cloned = Arc::clone(&wlts);
+            let txid_recipient = Arc::clone(&tx_id);
+            let ok =
+                handle_notification(notif, &cl, &sub_id, cloned, p_id, now, txid_recipient).await?;
+            Ok(ok)
+        });
+        match tokio::time::timeout(max_wait, notif_handler).await {
```
tompro, line 220: If you don't intend to listen for push notifications permanently you can just query for your message filter directly like:
```
client.fetch_events_from(
  relays (eg from config),
  filter (the filter as you have it for subscription),
  timeout,
)
// or:
client.fetch_events(filter, timeout) // using the clients relays
```

tompro: Timeout is similar to a timeout of an http request. From the method docs `This is an **auto-closing subscription** and will be closed automatically on 'EOSE'`, where EOSE is a message from the relay saying end of stream.

tompro (approved):

PR #83, opened 2025-10-22
Initial offline intermint

stefanbitcr:
Offline intermint for wallet
Multihop case consists of first exchanging eCash for substitute beta eCash, and then doing atomic HTLC exchange from there to the wallet mint.

### DIFF
--- a/crates/bcr-wallet-core/src/utils.rs
+++ b/crates/bcr-wallet-core/src/utils.rs
@@ -5,7 +5,6 @@ use bitcoin::secp256k1::{PublicKey, SECP256K1};
 use cashu::{BlindSignature, Proof};
 use cdk::Error as CdkError;
 // ----- local imports
-use crate::mint::MintConnector;
 use crate::mint::SecretlessProof;
 // ----- end imports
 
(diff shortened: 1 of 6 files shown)


crates/bcr-wallet-core/src/wallet.rs
```diff
@@ -455,10 +498,43 @@ where
...
+        // Ephemeral P2PK secret
+        let wallet_pk = cashu::SecretKey::generate();
+
+        let (secretless, secrets) =
+            proofs_to_secretless(alpha_id, substitute_client, proofs).await?;
+        let hash_locks: Vec<Sha256> = secrets
+            .iter()
+            .map(|secret| Sha256::hash(&secret.to_bytes()))
+            .collect();
+        let mut beta_proofs = substitute_client
```
zupzup, line 517: Are these guaranteed (under all circumstances / error cases) to come back in the exact same order as the hash locks that were sent?

zupzup (approved):
LGTM 👍

mtbitcr (approved):

crates/bcr-wallet-core/src/utils.rs
```diff
@@ -1,18 +1,80 @@
...
 // ----- local imports
-
+use crate::mint::MintConnector;
+use crate::mint::SecretlessProof;
 // ----- end imports
 
-#[cfg(test)]
+type CdkResult<T> = std::result::Result<T, cdk::Error>;
+
+pub async fn proofs_to_secretless(
```
codingpeanut157, line 13: Is this used for external proofs or for ours as well?
what if we store the blinded signatures together with the proofs?

crates/bcr-wallet-core/src/utils.rs
```diff
@@ -1,18 +1,80 @@
...
+    alpha_id: PublicKey,
+    substitute_client: &crate::mint::HttpClientExt,
```
codingpeanut157: I think you should take a `&dyn crate::mint::MintConnector` here

codingpeanut157 (approved):

crates/bcr-wallet-core/src/utils.rs
stefanbitcr: For external alpha eCash to be converted to Beta eCash, this way we don't reveal the secret to the substitute Beta, but do the blinded signature.

> what if we store the blinded signatures together with the proofs?

Which use case were you thinking of ?

crates/bcr-wallet-core/src/wallet.rs
stefanbitcr: Good question, they should, we can add a check if they indeed do rather then assuming so, when adding the corresponding proof-secret preimage.

Worst case we error, avoid revealing the preimages, and the beta proofs expire.

crates/bcr-wallet-core/src/utils.rs
codingpeanut157: ah no sorry
my mistake
never mind

PR #86, opened 2025-10-28
Refactor to use fingerprints

stefanbitcr:
Clowder updated to use fingerprints, these changes are to match the request format expected for the offline intermint exchange.

### DIFF
--- a/crates/bcr-wallet-core/src/mint.rs
+++ b/crates/bcr-wallet-core/src/mint.rs
@@ -4,7 +4,7 @@ use std::str::FromStr;
 use async_trait::async_trait;
 use bitcoin::hashes::sha256::Hash as Sha256;
 use bitcoin::secp256k1::PublicKey;
-use cashu::{BlindSignature, Proof, ProofDleq};
+use cashu::{Proof, ProofDleq};
 use cdk::Error as CdkError;
 use serde::{Deserialize, Serialize};
 // ----- local imports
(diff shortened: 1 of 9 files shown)


zupzup (approved):
LGTM 👍

crates/bcr-wallet-core/src/mint.rs
```diff
@@ -42,15 +42,18 @@ pub struct PublicKeyResponse {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct SecretlessProof {
-    pub signature: BlindSignature,
-    pub y: bitcoin::secp256k1::PublicKey,
-    pub dleq: ProofDleq,
+pub struct ProofFingerprint {
```
codingpeanut157, line 45: I have something similar in [here](https://github.com/BitcreditProtocol/bcr-common/blob/ce31d27b382393ec0b1c4d486e0a0a7c275a0d20/src/wire/keys.rs#L60)

maybe in the future we can merge the two and remove duplicates

codingpeanut157 (approved):
I am ok with duplicating structures in this phase where we need to run fast to get to the next checkpoint

In the future we definitely need to move more stuff into bcr-common and have them shared among wallet, wildcat and clowder

crates/bcr-wallet-core/src/mint.rs
stefanbitcr: Indeed, its the idea to reuse them for the offline intermint exchange. They are missing the dleq and witness so this is the first step.

mtbitcr (approved):

Issue #92, opened 2025-11-12
Revisit factory pattern


stefanbitcr:
The purse takes a trait object to create a new wallet. It might be beneficial to examine using a WalletFactory that can instantiate wallets (HD wallets ?)

Similarly when the wallet needs to query a distant mint via mint url, it would be beneficial to have a factory that can instantiate one based on the mint url, this would also be necessary when the betas change dynamically, as currently the beta clients and urls are determined prior to wallet creation.

zupzup:
Once we implement full support for multiple wallets in #170, this becomes more straightforward, since we'll have all the infrastructure needed to create / delete wallets on-demand.

zupzup:
architecture changed, multi wallets now

PR #105, opened 2025-11-18
[core] replace clowder_models with bcr_common::wire::clowder

codingpeanut157:


mtbitcr (approved):

crates/bcr-wallet-core/src/app.rs
```diff
@@ -390,6 +389,7 @@ pub async fn wallet_check_received_payment(
     Ok(tx_id)
 }
 
+// TODO: this needs fix as a mutex is held through await point
```
zupzup, line 392: This seems to happen at many more places (basically any async function on wallet where the RwLock is held across await, no?). 🤔

Cargo.toml
```diff
@@ -11,7 +11,7 @@ members = ["crates/bcr-wallet-core", "crates/bcr-wallet-lib"]
 [workspace.dependencies]
 anyhow = { version = "1" }
 async-trait = { version = "0.1" }
-bcr-common = {git = "https://github.com/BitcreditProtocol/bcr-common.git", branch = "master"}
+bcr-common = {git = "https://github.com/BitcreditProtocol/bcr-common.git", branch = "peanut/clowder_1"}
```
zupzup: I imagine this will be changed to `master` or whatever once the bcr-common thing is changed

zupzup (approved):
LGTM 👍

crates/bcr-wallet-core/src/app.rs
codingpeanut157: yes
it's a workaround introduced to get faster to intermint-exchange.
We have already an issue where we review the current design https://github.com/BitcreditProtocol/Wallet-Core/issues/92

with the new design we can remove, all the `(&mut self` functions from the wallet, and thus get rid of the mutexes wrapping the wallets

stefanbitcr (approved):

PR #115, opened 2025-12-15
Pay by Token and Fix Payment and Transaction Status

zupzup:
* Add `Pay by Token`, which creates and returns a debit, or credit token to be sent via QR code
* Fix PaymentStatus and TransactionStatus for the different payment methods
* Set unit and amount in payment preparation (they were missing/wrong in the summary)
* Fix/Clean up CLI display of transactions

### DIFF
--- a/crates/bcr-wallet-core/src/wallet.rs
+++ b/crates/bcr-wallet-core/src/wallet.rs
@@ -4,7 +4,10 @@ use crate::{
     error::{Error, Result},
     purse::{self},
     sync,
-    types::{self, MeltSummary, PaymentSummary, RedemptionSummary, SendSummary, WalletConfig},
+    types::{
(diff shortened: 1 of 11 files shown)


mtbitcr (approved):

crates/bcr-wallet-core/src/wallet.rs
```diff
@@ -978,6 +1007,88 @@ where
...
+                    quote_id: None,
+                };
+                let tx_id = self.tx_repo.store_tx(partial_tx).await?;
+                Ok((tx_id, None))
+            }
+            PaymentType::Token => {
+                let (proofs, token) = if unit == self.credit.unit() {
+                    let p = self
+                        .credit
+                        .send_proofs(
```
stefanbitcr, line 1039: ✅

stefanbitcr (approved):
👍

PR #116, opened 2025-12-16
List Transactions, Remove SettingsDB, Restructure App State Config

zupzup:
* Add endpoint for listing full transactions (instead of only being able to list ids and loading each one separately), returns Transactions sorted by timestamp descending
* Move Config outwards to the library caller from the SettingsDB
  * @mtbitcr Here we need to make a decision on whether we want 1 mnemonic for all (nostr client, wallet), or several and how initialization should work

### DIFF
--- a/crates/bcr-wallet-core/src/wallet.rs
+++ b/crates/bcr-wallet-core/src/wallet.rs
@@ -123,6 +123,7 @@ pub trait TransactionRepository: sync::SendSync {
     #[allow(dead_code)]
     async fn delete_tx(&self, tx_id: TransactionId) -> Result<()>;
     async fn list_tx_ids(&self) -> Result<Vec<TransactionId>>;
+    async fn list_txs(&self) -> Result<Vec<Transaction>>;
     async fn update_metadata(
         &self,
         tx_id: TransactionId,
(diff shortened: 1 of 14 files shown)


mtbitcr (approved):

crates/bcr-wallet-core/src/wallet.rs
```diff
@@ -123,6 +123,7 @@ pub trait TransactionRepository: sync::SendSync {
     #[allow(dead_code)]
     async fn delete_tx(&self, tx_id: TransactionId) -> Result<()>;
     async fn list_tx_ids(&self) -> Result<Vec<TransactionId>>;
+    async fn list_txs(&self) -> Result<Vec<Transaction>>;
```
codingpeanut157, line 126: the original idea behind `list_tx_ids` was to have some sort of  "loading while scrolling" effect without overwhelming the frontend with a huge pile of transactions objects.

Rather than doing classic paging with number of pages or indexes, I return the full list of txid (long list of compact items) to the front-end which then queries the ids as the user scrolls.

codingpeanut157 (approved):

crates/bcr-wallet-core/src/wallet.rs
zupzup: Yeah, it's still there - this was just a request for now to simplify things.

In the future this will be a filterable & paged endpoint.

Then the frontend can decide what fits best for which screen

PR #118, opened 2025-12-18
Dont store mnemonic

zupzup:
* Removes `mnemonic` from persistence layer
* Calculates public key and checks, if the existing wallet matches with the given mnemonic that way
* Only keeps mnemonic / seed in-memory, it's never stored

### DIFF
--- a/crates/bcr-wallet-core/src/lib.rs
+++ b/crates/bcr-wallet-core/src/lib.rs
@@ -16,7 +16,7 @@ use cashu::{CurrencyUnit, KeySetInfo, MintInfo, MintUrl};
 use cdk::wallet::{MintConnector as MintCon, types::TransactionId};
 use chrono::Utc;
 use error::{Error, Result};
-use secp256k1::{Keypair, SECP256K1, SecretKey};
+use secp256k1::{Keypair, SECP256K1};
 use std::{
     collections::{HashMap, HashSet},
     str::FromStr,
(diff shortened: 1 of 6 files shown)


stefanbitcr (approved):

crates/bcr-wallet-core/src/lib.rs
```diff
@@ -745,11 +748,26 @@ async fn build_wallet(
...
         beta_clients,
         Box::new(|url| Box::new(crate::mint::HttpClientExt::new(url))),
         same_mint_safe_mode,
     )
     .await?;
     Ok(new_wallet)
 }
+
+// converts a mnemonic to a secp256k1 keypair
+fn keypair_from_mnemonic(mnemonic: &bip39::Mnemonic) -> Result<Keypair> {
```
codingpeanut157: I think you can rule out the chances of an error here.
```rust
fn keypair_from_mnemonic(mnemonic: &bip39::Mnemonic) -> Keypair {
    let seed = seed_from_mnemonic(mnemonic);
    let (key, _) = seed.split_at(secp256k1::constant::SECRET_KEY_SIZE); // panics if seed.len() < SECRET_KEY_SIZE
    let keypair = Keypair::from_seckey_slice(SECP256K1, key).expect("key to be correct size");
    keypair
}
```

codingpeanut157 (approved):

PR #126, opened 2026-01-13
Implement Mint and Melt

zupzup:
* Removed `wallet_check_pending_melts` - since onchain melts execute immediately
* Add mint and melt
    * Add `wallet_prepare_melt` - prepares a melt, returns a payment summary - calling `wallet_pay` with the `request id` executes the melt
    * Add optional `btc_tx_id` to `Transaction` - the Bitcoin transaction ID (e.g. from a melt operation)
    * Add optional `quote_id` to `Transaction` - the Mint quote ID (e.g. from a mint operation)
    * Add `wallet_mint` -  creates a mint request for the given amount, returns a mint summary, with the amount and BTC address to pay to
    * Add `wallet_check_pending_mints` - checks the open mint requests and attempts to mint them, if they were paid (Also called during the regular job runs)

### DIFF
--- a/crates/bcr-wallet-core/src/pocket/debit.rs
+++ b/crates/bcr-wallet-core/src/pocket/debit.rs
@@ -1,17 +1,18 @@
 use crate::{
     MintConnector,
     error::{Error, Result},
+    keypair_from_seed,
     pocket::*,
     restore,
-    types::{MeltSummary, SendSummary},
+    types::{MeltSummary, MintSummary, SendSummary},
     wallet,
 };
 use async_trait::async_trait;
+use bcr_common::wire::{melt as wire_melt, mint as wire_mint};
 use cashu::{
     Amount, CurrencyUnit, KeySet, KeySetInfo, amount::SplitTarget, nut00 as cdk00, nut01 as cdk01,
-    nut05 as cdk05, nut23 as cdk23,
+    nut05 as cdk05,
 };
-use cdk::Error as CdkError;
 use std::{
     collections::HashMap,
     sync::{Arc, Mutex},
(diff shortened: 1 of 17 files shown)


crates/bcr-wallet-core/src/pocket/debit.rs
```diff
@@ -68,6 +81,42 @@ impl Pocket {
         }
     }
 
+    pub fn generate_blinds(
```
stefanbitcr: nitpick:
There is a cashu Premint struct that contains the (blind,secret,r,amount)
and PreMintSecrets containing keyset_id, `Vec<Premints>`

stefanbitcr (approved):

crates/bcr-wallet-core/src/pocket/debit.rs
```diff
@@ -68,6 +81,42 @@ impl Pocket {
...
+    pub fn generate_blind(
+        kid: cashu::Id,
+        amount: cashu::Amount,
+    ) -> Result<(
+        cashu::BlindedMessage,
+        cashu::secret::Secret,
+        cashu::SecretKey,
+    )> {
+        let secret = cashu::secret::Secret::new(hex::encode(rand::random::<[u8; 32]>()));
+        let (b_, r) = cashu::dhke::blind_message(secret.as_bytes(), None)?;
```
stefanbitcr: PreMintSecrets::random for generation

Although elsewhere we generate the secrets for the blinds deterministically (Nut 13) by increment a keyset counter, as that allows (Nut 09) recovery

Recovery is very slow for large amount of eCash swapped (old wallets) so I'm not sure how much value there is to follow it though.

zupzup: Oh, this is very nice, thx! - We can also use this much simpler method in ebill then :+1:

PR #129, opened 2026-01-16
Refactor Rust CI workflow for clarity and updates

mtbitcr:
Updated Rust CI workflow with improved steps and comments.


.github/workflows/rust.yml
mtbitcr: Still needed?

zupzup: No, I think both can be removed

zupzup (approved):

cleot (approved):

PR #139, opened 2026-01-21
Use clowder client from BCR Common

stefanbitcr:
Current endpoints drifted as the original implementation was not with Wildcat


mtbitcr (approved):

codingpeanut157 (approved):

zupzup (approved):
LGTM 👍

PR #144, opened 2026-01-22
Fix intermint exchange

zupzup:
* Fix intermint exchange using adaptations from Clowder demowallet
* Add carol
* Add keypair to wallet in-memory object instead of just pub key

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.7.6
+
+* Fix intermint exchange
+
 # 0.7.5
 
 * Adapt to new Clowder URLs
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace.package]
-version = "0.7.5"
+version = "0.7.6"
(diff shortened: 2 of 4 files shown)


stefanbitcr (approved):
Minor concern if the wallet ends up reusing the public key it loses some privacy

zupzup:
> Minor concern if the wallet ends up reusing the public key it loses some privacy

Oh, I thought we need the wallet pub key instead of the ephemeral one now - I guess I misunderstood. Then I'll revert that part and keep the ephemeral key, right?

stefanbitcr:
> > Minor concern if the wallet ends up reusing the public key it loses some privacy
> 
> Oh, I thought we need the wallet pub key instead of the ephemeral one now - I guess I misunderstood. Then I'll revert that part and keep the ephemeral key, right?

Either is fine, functionally they shouldn't be different.

As long as you can add a signature on the received Beta proofs with the same key used in the initial request (added to the exchange path).

mtbitcr (approved):

PR #158, opened 2026-01-29
Offline Pay by Token, Migrate Rabid and other Changes

zupzup:
* Add `clowder_id`, `betas` and `mint_keysets` to `WalletConfig` (breaking DB change)
* Improve Offline functionality and performance
    * Clowder ID is fetched at wallet initialization and cached in DB
    * Betas are fetched at wallet initialization and cached in DB
    * Mint keysets are fetched at wallet initialization and cached in DB / refetched on-demand
* We always initialize Credit Sat Pocket now with `crsat`, even if the Mint doesn't have a credit keyset
* Add endpoints `wallet_mint_is_rabid` and `wallet_mint_is_offline` to check whether a wallet mint 
* Removed `purse_migrate_rabid` from daily jobs - it now has to be called directly and returns a map of migrated wallets with their new mints
* Removed the check for `default_mint_url` to have to match the wallet - it's just logged now
* Implement a hacky demo-version of `offline_pay_by_token`, where the wallet can create a token even if it's alpha mint is offline

### DIFF
--- a/crates/bcr-wallet-core/src/pocket/credit.rs
+++ b/crates/bcr-wallet-core/src/pocket/credit.rs
@@ -354,10 +354,10 @@ impl wallet::Pocket for Pocket {
         let mut premints = HashMap::from([(active_info.id, premint_secrets)]);
         let keysets = HashMap::from([(active_info.id, active_keyset)]);
 
-        let mut blinds: Vec<cdk00::BlindedMessage> = Vec::new();
-        for premint in premints.values() {
-            blinds.extend(premint.blinded_messages());
-        }
+        let blinds: Vec<cdk00::BlindedMessage> = premints
+            .values()
+            .flat_map(|premint| premint.blinded_messages())
+            .collect();
 
         // swap
         let request = cdk03::SwapRequest::new(proofs, blinds);
@@ -370,17 +370,20 @@ impl wallet::Pocket for Pocket {
                 .and_modify(|v| v.push(signature.clone()))
                 .or_insert_with(|| vec![signature]);
         }
+
+        let mut current_amount = Amount::ZERO;
         // only collect sending proofs - change is discarded for now
         for (kid, signatures) in signatures.into_iter() {
             let premint = premints.remove(&kid).expect("premint should be here");
             let keyset = keysets.get(&kid).expect("keyset should be here");
             let mut unblinded_proofs = unblind_proofs(keyset, signatures, premint);
             unblinded_proofs.sort_by_key(|proof| std::cmp::Reverse(proof.amount));
-            let mut current_amount = Amount::ZERO;
             for proof in unblinded_proofs.into_iter() {
                 if current_amount + proof.amount <= send_amount {
                     current_amount += proof.amount;
                     swapped_proofs.push(proof);
+                } else {
+                    break;
                 }
             }
         }
--- a/crates/bcr-wallet-core/src/pocket/credit.rs
+++ b/crates/bcr-wallet-core/src/pocket/credit.rs
@@ -377,16 +377,14 @@ impl wallet::Pocket for Pocket {
         }
 
         let mut current_amount = Amount::ZERO;
-        // only collect sending proofs - change is discarded for now
+        // only collect sending proofs, so we can take all - change is discarded for now
         for (kid, signatures) in signatures.into_iter() {
             let premint = premints.remove(&kid).expect("premint should be here");
             let keyset = keysets.get(&kid).expect("keyset should be here");
             let unblinded_proofs = unblind_proofs(keyset, signatures, premint);
             for proof in unblinded_proofs.into_iter() {
-                if current_amount + proof.amount <= send_amount {
-                    current_amount += proof.amount;
-                    swapped_proofs.push(proof);
-                }
+                current_amount += proof.amount;
+                swapped_proofs.push(proof);
             }
         }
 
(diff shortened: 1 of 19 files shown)


mtbitcr (approved):

crates/bcr-wallet-core/src/pocket/credit.rs
```diff
@@ -293,14 +293,103 @@ impl wallet::Pocket for Pocket {
...
+        // create premints - no counter etc., since we're not persisting them anyway
+        let premint_secrets = cashu::PreMintSecrets::random(
+            active_info.id,
+            total_amount,
+            &SplitTarget::Values(splits),
+        )?;
+        let mut premints = HashMap::from([(active_info.id, premint_secrets)]);
+        let keysets = HashMap::from([(active_info.id, active_keyset)]);
+
+        let mut blinds: Vec<cdk00::BlindedMessage> = Vec::new();
```
stefanbitcr: nitpitck: could also map and collect ?

crates/bcr-wallet-core/src/pocket/credit.rs
```diff
@@ -293,14 +293,103 @@ impl wallet::Pocket for Pocket {
...
+
+        let mut signatures: HashMap<cashu::Id, Vec<cdk00::BlindSignature>> = HashMap::new();
+        for signature in response.signatures {
+            signatures
+                .entry(signature.keyset_id)
+                .and_modify(|v| v.push(signature.clone()))
+                .or_insert_with(|| vec![signature]);
+        }
+        // only collect sending proofs - change is discarded for now
+        for (kid, signatures) in signatures.into_iter() {
```
stefanbitcr, line 381: Can't we also iterate here over only send_splits len since the premints should be in order ?

Rather than trying to fetch the correct amount from the signatures again. If split target is values and it matches the same amount it just returns and uses those again.

stefanbitcr (approved):
Nice!

crates/bcr-wallet-core/src/pocket/credit.rs
```diff
@@ -293,14 +293,103 @@ impl wallet::Pocket for Pocket {
...
+            let premint = premints.remove(&kid).expect("premint should be here");
+            let keyset = keysets.get(&kid).expect("keyset should be here");
+            let mut unblinded_proofs = unblind_proofs(keyset, signatures, premint);
+            unblinded_proofs.sort_by_key(|proof| std::cmp::Reverse(proof.amount));
+            let mut current_amount = Amount::ZERO;
+            for proof in unblinded_proofs.into_iter() {
+                if current_amount + proof.amount <= send_amount {
+                    current_amount += proof.amount;
+                    swapped_proofs.push(proof);
+                }
```
codingpeanut157: maybe a break to early exit this loop?

crates/bcr-wallet-core/src/pocket/debit.rs
```diff
@@ -401,6 +401,92 @@ impl wallet::Pocket for Pocket {
...
+            let premint = premints.remove(&kid).expect("premint should be here");
+            let keyset = keysets.get(&kid).expect("keyset should be here");
+            let mut unblinded_proofs = unblind_proofs(keyset, signatures, premint);
+            unblinded_proofs.sort_by_key(|proof| std::cmp::Reverse(proof.amount));
+            let mut current_amount = Amount::ZERO;
+            for proof in unblinded_proofs.into_iter() {
+                if current_amount + proof.amount <= send_amount {
+                    current_amount += proof.amount;
+                    swapped_proofs.push(proof);
+                }
```
codingpeanut157: maybe a break to early exit this loop?

codingpeanut157 (approved):

crates/bcr-wallet-core/src/pocket/credit.rs
zupzup: Yeah, I also just saw there's a bug because the `current_amount` is in the inner loop, but should be across all signatures and proofs, right? (Guess I was lucky in the test)

I added a break once the target amount is reached, which should come out to the same, since the `send_splits` are used first, right? While staying more flexible if the PreMintSecrets definition changes to use a different split target.

I'll likely remove this whole code path anyway when reworking the pockets a bit to separate persistence from swapping logic to be a bit more flexible.

stefanbitcr: Yeah the grouping by keyset and then sorting is the problematic part since the insert order isn't defined ?. We might miss our target if the next one is greater?

zupzup: Ah right. So basically I could just take the first `send_split.len()` signatures and then no matter how I group and sort, it has to match my target (send_amount) when I go through all proofs, right?

stefanbitcr: I think they have to the same original signatures as we know those first len amount are exactly the send set.
So you only take those (first send_split.len()), then do the grouping and whatever else after. 

The rest is the change amounts we don't use

crates/bcr-wallet-core/src/pocket/credit.rs
```diff
@@ -376,17 +381,21 @@ impl wallet::Pocket for Pocket {
         for (kid, signatures) in signatures.into_iter() {
             let premint = premints.remove(&kid).expect("premint should be here");
             let keyset = keysets.get(&kid).expect("keyset should be here");
-            let mut unblinded_proofs = unblind_proofs(keyset, signatures, premint);
-            unblinded_proofs.sort_by_key(|proof| std::cmp::Reverse(proof.amount));
+            let unblinded_proofs = unblind_proofs(keyset, signatures, premint);
             for proof in unblinded_proofs.into_iter() {
                 if current_amount + proof.amount <= send_amount {
                     current_amount += proof.amount;
                     swapped_proofs.push(proof);
```
stefanbitcr: Can just take all proofs here as the send amount should match exactly (after taking only required sigs changes)

PR #167, opened 2026-02-11
Cleanup, Refactoring, DB Tests, Basic Crate Structure

zupzup:
First step of a big refactoring effort to get wallet on track for more stability in the futre:

* Split into multiple crates (basic version - core will be fleshed out more)
* Add DB Tests
* General cleanup

Next step will be using strong types (no strings, u64 etc.) everywhere possible, more tests, structural improvements of the Purse/Wallet/Pocket structure (locks etc.) and finally moving wallet-ffi into this repo.


mtbitcr (approved):

crates/bcr-wallet-core/src/utils.rs
```diff
@@ -1,355 +1,23 @@
...
-
-    let signatures = vec![signature.to_string()];
-
-    proof.witness = Some(cashu::Witness::HTLCWitness(HTLCWitness {
-        preimage: preimage.to_string(),
-        signatures: Some(signatures),
-    }));
 
-    Ok(())
+pub fn sum_proofs<'a, I>(proofs: I) -> cashu::Amount
```
codingpeanut157: there is a trait defined by cashu and implemented for a `Vec` and `Hashset` that might help you here

[cashu::nuts::nut00::ProofsMethods](https://docs.rs/cashu/0.13.1/cashu/nuts/nut00/trait.ProofsMethods.htmll)

crates/bcr-wallet-api/src/wallet.rs
```diff
@@ -53,7 +57,7 @@ impl SafeMode {
 /// trait that represents a single compartment in our wallet where we store proofs/tokens of the
 /// same currency emitted by the same mint
 #[async_trait]
-pub trait Pocket: sync::SendSync {
+pub trait Pocket: SendSync {
```
codingpeanut157, line 60: another good move to reduce the size of wallet file might be to move Pocket trait to `pocket/mod.rs`

crates/bcr-wallet-persistence/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+    async fn store(&self, wallet: WalletConfig) -> Result<()>;
+    async fn load(&self, wallet_id: &str) -> Result<WalletConfig>;
+    async fn delete(&self, wallet_id: &str) -> Result<()>;
+    async fn list_ids(&self) -> Result<Vec<String>>;
+}
+
+///////////////////////////////////////////// TransactionRepository
+#[cfg_attr(any(test, feature = "test-utils"), mockall::automock)]
+#[async_trait]
+pub trait TransactionRepository: SendSync {
```
codingpeanut157, line 59: nitpicking:
while I agree with moving the `Repository` traits to the `persistence/mod.rs` file, here you are moving the trait to a different crate.
That means you are making `wallet-persistence` a  dependency of `wallet-api` while it logically makes more sense to me to have the opposite. 
For instance, imagine I want to implement `TransactionRepository`  for my own persistence technology, with this change I have to import `wallet-persistence` with all the redb crates and such

codingpeanut157 (approved):

crates/bcr-wallet-persistence/src/lib.rs
zupzup: That's a good point - do you think moving the trait definitions into `bcr-wallet-core` would make more sense?

Otherwise, I could feature-gate the whole redb implementation, then by default you'd always only get the traits and if you want a concrete implementation (currently only redb), you can opt into that. Seems the better way on first glance :thinking:

crates/bcr-wallet-api/src/wallet.rs
zupzup: Yeah, I'll do that restructuring in the next PR :+1:

crates/bcr-wallet-persistence/src/lib.rs
codingpeanut157: if I were you I'd leave the Repository traits in the wallet-api crate, maybe in a separate module named `repository`
so you'd have
`repository::Transactions` and `repository::Proofs`, so on and so forth
and then have `wallet-persistence` importing `wallet-api` and implement those traits

PR #168, opened 2026-02-13
Restructure Purse and Wallet, Add Tests

zupzup:
Progress on the Refactoring Effort

* Move some types and utils to core, or where they belong
* Restructure purse/wallet/pocket and mint code
    * Mods for wallet, purse and mint
    * Move traits to their impls
    * Split up Wallet for structure
    * Move wallet specific functions directly to wallet, not going through purse
* Add Purse Tests
* Rework Errors structure
* Clean up outer types a bit

The third and last step (for now) will add some wallet test baseline (have to refactor the dyn ClowderMintConnector for that first), Strong types on the FFI bounday and finally moving wallet-ffi into this repo.


crates/bcr-wallet-api/src/pocket/mod.rs
```diff
@@ -1,19 +1,66 @@
...
-    wallet::SafeMode,
+    wallet::types::SafeMode,
 };
+use async_trait::async_trait;
 use bcr_common::cashu::{
     self, Amount, CurrencyUnit, KeySet, KeySetInfo, ProofsMethods, amount::SplitTarget,
     nut00 as cdk00, nut01 as cdk01, nut03 as cdk03, nut07 as cdk07,
 };
+use bcr_wallet_core::{SendSync, types::SendSummary};
 use bcr_wallet_persistence::PocketRepository;
```
codingpeanut157, line 12: `wallet-api` depending on `wallet-persistence` 😖 

the world is upside down 😱 

😆 

I bet you have programmed in C++ or Java in the past... 🤔

crates/bcr-wallet-api/src/purse/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+};
+use bcr_common::cashu::MintUrl;
+use bcr_wallet_core::types::WalletConfig;
+use bcr_wallet_persistence::{PurseRepository, redb::purse::PurseDB};
+use std::{collections::HashMap, sync::Arc};
+use tokio::sync::RwLock;
+
+pub struct Purse<Wlt> {
+    pub repo: Box<dyn PurseRepository>,
+    pub wallets: Arc<RwLock<Vec<Arc<RwLock<Wlt>>>>>,
```
codingpeanut157, line 13: not sure if this was already there or not
but given this is a library for end-user app, I am not sure an `RwLock` would improve things that much compared to a simple `Mutex`

crates/bcr-wallet-api/src/wallet/mod.rs
```diff
@@ -0,0 +1,830 @@
...
+    hashes::{Hash, sha256::Hash as Sha256},
+    secp256k1,
+};
+use nostr::{nips::nip59::UnwrappedGift, signer::NostrSigner};
+use nostr_sdk::nips::nip19::{FromBech32, Nip19Profile};
+use std::{collections::HashMap, str::FromStr, sync::Arc};
+use tokio::sync::Mutex;
+use uuid::Uuid;
+
+pub struct Wallet {
```
codingpeanut157, line 34: isn't this layout a bit inconsistent with `pocket`?
In pocket you declare the trait in the mod.rs and implementations in submodules.
Here you do the opposite

codingpeanut157 (approved):
minor changes regarding file and types layouts

crates/bcr-wallet-api/src/purse/mod.rs
zupzup: Quite the opposite, actually. Changing this to RwLock was the single biggest performance improvement, that got us from sluggish to responsive.

This wasn't really noticeable in the html harness, but very directly with the native app.

It's not that surprising, since a wallet and how the API is used by the native app (mostly small, sometimes concurrent reads) is very read-heavy and before, reads blocked each other

crates/bcr-wallet-api/src/wallet/mod.rs
zupzup: In pocket (which is structurally inherently different because of the two types) each part fits in it's own file.

Once I get to untangling the swapping code from persistence, I expect this will lead to changes there too. This is merely a first pass.

crates/bcr-wallet-api/src/pocket/mod.rs
zupzup: Not sure what you mean here 🤔 

Api (business logic + public api to ffi in this case) depending on the persistence layer, which in turn depends on core (types, utils, traits) seems pretty uneventful to me.

I just oriented myself by the Ebill crate structure, which developed over time and turned out to be practical, but I'm all ears for different approaches.

stefanbitcr (approved):

PR #173, opened 2026-02-20
Alpha and beta tx id for melts

zupzup:
* Inflated because of Cargo.lock check-in
* Adapts to the changes in https://github.com/BitcreditProtocol/bcr-common/pull/64 (bcr-common dep has to be changed after it's merged)


stefanbitcr (approved):
Nice ! Thanks

PR #177, opened 2026-03-10
Update Minting with latest BCR Common

stefanbitcr:
* Update Wallet to use latest BCR Common type for minting 
* Check signature from mint

Out of scope:
* Protest

Todo:

- [x] Switch to merged BCR Common

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 # 0.8.1
 
+* Update minting flow - breaking database change for storing premint secrets during minting
 * Persist alpha_tx_id and beta_tx_id for melts as per https://github.com/BitcreditProtocol/Clowder/pull/207 (breaking DB change)
     * replace `btc_tx_id` with optional `melt_tx` struct on `Transaction` (breaking API change)
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -301,7 +301,7 @@ checksum = "2af50177e190e07a26ab74f8b1efbfe2ef87da2116221318cb1c2e82baf7de06"
 [[package]]
 name = "bcr-common"
 version = "0.7.0"
-source = "git+https://github.com/BitcreditProtocol/bcr-common.git?branch=stefan%2Fcommitment_borsch#5d308d5da6be31fb689e963a41acfc58cdcab611"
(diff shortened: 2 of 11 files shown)


zupzup (approved):
Nice, thx for taking care of the Wallet side of this! 🚀 

Could you please add an entry in `CHANGELOG.md` as well, mentioning that it's a breaking DB change (premints)

zupzup:
@stefanbitcr Are the Clowder/Wildcat changes for this already deployed? If so, I would cut a new release once this is merged so minting continues to work on clowder-dev

mtbitcr (approved):

stefanbitcr:
> @stefanbitcr Are the Clowder/Wildcat changes for this already deployed? If so, I would cut a new release once this is merged so minting continues to work on clowder-dev

None are yet (hopefully) deployed, I will update all repositories so that compatibility is maintained

codingpeanut157 (approved):
maybe it's worth rebasing after this 
https://github.com/BitcreditProtocol/bcr-common/pull/78
gets merged

The main change is in the responseBody: I applied the same pattern we applied in other places where a signed message is 
```
pub struct Payload {
...
}

pub struct Message {
  content: String, // base64, borsh serialized Payload
  signature: schnorr::Signature,
}
```
because json can potentially reorder fields and elements in an array, we can't send the Body as a nested structure otherwise on the other side the signature verification can potentially fail.
We have to send the closest representation of the content we signed

You can then use `schnorr_verify_b64` and `deserialize_borsh_msg` from `bcr-common::core::signature`

stefanbitcr:
@codingpeanut157  

> we can't send the Body as a nested structure otherwise on the other side the signature verification can potentially fail.

We could, as long as both sides serialize the data (body : ), say with borsch, to binary canonically, then the problem is solved as well since you'd get a deterministic hash. (+blinded message handling)

But this looks simpler to keep consistency and make our signing procedure more explicit.

codingpeanut157:
@stefanbitcr 

> We could, as long as both sides serialize the data (body : ), say with borsch

partially true.
You are right, a struct can beserialized and deserialized with json and then with borsh and everything should work fine.
It doesn't work with arrays though.
If json shuffles the elements ( the standard does not force to maintain order ) , then borsh can't do anything about it.

codingpeanut157 (approved):

PR #180, opened 2026-03-23
Add initial protest

stefanbitcr:


crates/bcr-wallet-ffi/src/frb_generated.rs
```diff
@@ -26,7 +26,7 @@
 // Section: imports
```
zupzup, line 27: I guess this change is from `cargo fmt` - shouldn't matter.

Did you run `flutter_rust_bridge_codegen generate` (the flow is documented in the readme)

Also, do we need to add this to the ffi layer? (i.e. - does the user have to be able to trigger it?).

You can orient yourselves at the stuff in `bcr-wallet-ffi`.

Once you add stuff there, you have to run `flutter_rust_bridge_codegen generate` to generate the bindings.

Sometimes, if you add new structs, you have to `generate` once, before the Rust stuff compiles.

We can also go over that together if you want ✌️

zupzup (approved):
LGTM 👍  - see my comment regarding ffi

mtbitcr (approved):
@stefanbitcr forgot to include changes in wallet FFI. @zupzup will teach you (check READ.ME)

zupzup (approved):
Nice! 💪

PR #188, opened 2026-04-21
Melt Commitments

stefanbitcr:
TODO

- [x] Rebase on latest master

### DIFF
--- a/crates/bcr-wallet-api/src/pocket/debit.rs
+++ b/crates/bcr-wallet-api/src/pocket/debit.rs
@@ -21,22 +21,6 @@ use std::{
 };
 use uuid::Uuid;
 
-/// Sign the preimage of a base64-encoded content string with the ephemeral keypair.
-/// Used for protest request wallet_signatures.
-fn sign_content_b64(
-    content_b64: &str,
-    ephemeral_keypair: &secp256k1::Keypair,
-) -> Result<secp256k1::schnorr::Signature> {
-    use bitcoin::base64::{Engine, engine::general_purpose::STANDARD};
-    use bitcoin::hashes::{Hash, sha256};
-    let content_bytes = STANDARD
-        .decode(content_b64)
-        .map_err(|e| Error::MintingError(format!("invalid base64 content: {e}")))?;
-    let digest = sha256::Hash::hash(&content_bytes);
-    let msg = secp256k1::Message::from_digest(digest.to_byte_array());
-    Ok(secp256k1::SECP256K1.sign_schnorr(&msg, ephemeral_keypair))
-}
-
 #[async_trait]
 pub trait DebitPocketApi: super::PocketApi {
     /// Reclaim the proofs for the given ys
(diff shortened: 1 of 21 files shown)


crates/bcr-wallet-api/src/wallet/api.rs
```diff
@@ -431,17 +474,15 @@ impl WalletApi for super::Wallet {
                     TRANSACTION_STATUS_METADATA_KEY.to_owned(),
                     TransactionStatus::Settled.to_string(),
                 );
-
                 if let Some(alpha_tx_id) = btc_tx_id.alpha_txid {
                     metadata.insert(
                         BTC_ALPHA_TX_ID_TYPE_METADATA_KEY.to_owned(),
```
zupzup, line 439: maybe also `bcr_wallet_core::types::BTC_ALPHA_TX_ID_TYPE_METADATA_KEY` for consistency with the BETA_TX_ID key below

crates/bcr-wallet-persistence/src/redb/mintmelt.rs
```diff
@@ -85,8 +85,10 @@ struct MintEntry {
     kid: cashu::Id,
     content: String,
     commitment: bitcoin::secp256k1::schnorr::Signature,
+    ephemeral_secret: Vec<u8>,
 }
 
+#[allow(clippy::too_many_arguments)]
```
zupzup: Is this needed? 🤔 we have `too-many-arguments-threshold=200` in `.clippy.toml`, which should cover this nonsense for the whole workspace

crates/bcr-wallet-api/src/pocket/debit.rs
```diff
@@ -8,23 +8,35 @@ use async_trait::async_trait;
...
 use bitcoin::secp256k1;
 use std::{
     collections::HashMap,
     sync::{Arc, Mutex},
 };
 use uuid::Uuid;
 
+/// Sign the preimage of a base64-encoded content string with the ephemeral keypair.
+/// Used for protest request wallet_signatures.
+fn sign_content_b64(
```
zupzup: Based on the current structure, I'd move this to `pocket/mod.rs`, to the other utils.

Also, I think the `ephemeral_keypair` can just be `keypair`, since this could be used for any b64 signing.

codingpeanut157 (approved):

crates/bcr-wallet-api/src/pocket/debit.rs
```diff
@@ -557,32 +577,54 @@ impl DebitPocketApi for Pocket {
...
+            self.pdb.as_ref(),
+            &client,
+            Some(info.id),
+            swap_config.clone(),
+        )
+        .await?;
+
+        let proofs: Vec<cashu::Proof> = sending_proofs.values().cloned().collect();
+
+        let parsed_address: bitcoin::Address<bitcoin::address::NetworkUnchecked> = address
```
zupzup: I guess this check could happen at the very start of the method, since if the address is invalid, we don't need to do any of the rest (which includes some expensive http requests etc.)

zupzup (approved):
LGTM 👍

mtbitcr (approved):

PR #195, opened 2026-04-27
Don't ignore keysets with fees

zupzup:
* Until now, we assumed keysets to have 0 fees, which changes now
* Actual support for calculating fees will be implemented in a separate PR

### DIFF
--- a/crates/bcr-wallet-api/src/wallet/mod.rs
+++ b/crates/bcr-wallet-api/src/wallet/mod.rs
@@ -147,14 +147,14 @@ impl Wallet {
             // This can be revised based on some criteria ?
             let alpha_keysets = substitute_client.get_alpha_keysets(alpha_id).await?;
 
-            // The endpoint only returns active keysets and Clowder/Wildcat don't have fees
+            // The endpoint only returns active keysets
             let intermint_alpha_infos: Vec<cashu::KeySetInfo> = alpha_keysets
                 .iter()
                 .map(|keyset| cashu::KeySetInfo {
                     id: keyset.id,
                     unit: keyset.unit.clone(),
                     active: true,
-                    input_fee_ppk: 0,
+                    input_fee_ppk: 0, // TODO: get fees from the clowder keyset
                     final_expiry: keyset.final_expiry,
                 })
                 .collect();
--- a/crates/bcr-wallet-api/src/pocket/debit.rs
+++ b/crates/bcr-wallet-api/src/pocket/debit.rs
@@ -144,7 +144,7 @@ impl Pocket {
     fn find_active_keysetid(&self, keysets_info: &[KeySetInfo]) -> Result<cashu::KeySetInfo> {
         let active_info = keysets_info
             .iter()
-            .find(|info| info.unit == self.unit && info.active && info.input_fee_ppk == 0);
+            .find(|info| info.unit == self.unit && info.active);
         let Some(active_info) = active_info else {
             return Err(Error::NoActiveKeyset);
         };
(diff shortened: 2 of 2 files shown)


mtbitcr (approved):

crates/bcr-wallet-api/src/wallet/mod.rs
```diff
@@ -147,14 +147,14 @@ impl Wallet {
...
-            // The endpoint only returns active keysets and Clowder/Wildcat don't have fees
+            // The endpoint only returns active keysets
             let intermint_alpha_infos: Vec<cashu::KeySetInfo> = alpha_keysets
                 .iter()
                 .map(|keyset| cashu::KeySetInfo {
                     id: keyset.id,
                     unit: keyset.unit.clone(),
                     active: true,
-                    input_fee_ppk: 0,
+                    input_fee_ppk: 0, // TODO: get fees from the clowder keyset
```
zupzup, line 157: @stefanbitcr @codingpeanut157 

Currently, there's only `get_alpha_keysets` exposed via bcr-common (which gets the /foreign/keys), but doesn't include the `input_fee_ppk` field.

I see that Clowder also has a `/foreign/{alpha}/keyset/{keyset_id}` endpoint, where I could fetch this info for each keyset.

But would it maybe be possible to have an endpoint like `/v1/core/keysets` (list_keysets_info), where I could get keysetinfos with a filter (e.g. for active)?

Otherwise I'd have to make a request for each keyset here to get this info.

stefanbitcr: @zupzup 

keyset infos should actually only be for active keysets and this is what Cashu does

zupzup: @stefanbitcr 

Yeah, my point is rather about efficiency. To get to the keyset fee, I'd now have to first call `get_active_keysets` and in a second step, call `get_keyset_info` (which doesn't exist yet for foreign in bcr-common) for each of those keysets.

So I was just wondering if it would be possible to have an endpoint that returns a `Vec<cashu::KeySetInfo>` (which exists on core-service on /v1/core/keysets, can be filtered). It would be fine if it only returned active keysets, but it would save a lot of additional http traffic from the wallet.

codingpeanut157: The need to get the keyset info for the whole set of keysets of a foreign mint should be only in the case of foreign online exchange.

On the other hand I see this function is used in the more "popular" cases

- [Wallet::receive_proofs](https://github.com/BitcreditProtocol/Wallet-Core/blob/master/crates/bcr-wallet-api/src/wallet/api.rs#L715) 
- [Wallet::receive_token](https://github.com/BitcreditProtocol/Wallet-Core/blob/7ffb486e8f5f14597c6d9387ffa0320dc5edbf66/crates/bcr-wallet-api/src/wallet/mod.rs#L586)

did the protocol change recently?

codingpeanut157 (approved):

crates/bcr-wallet-api/src/wallet/mod.rs
zupzup: @codingpeanut157 Yes, this code path is in the intermint exchange path and there it gets the alpha keysets from the /foreign/keys endpoint, however it defaulted fees to 0 (since we didnt have fees so far), but now I guess I have to use the actual fees of the keyset here, right?

and in the default same-mint case, it just defaults to the alpha keysets

```
        if mint_url == self.client.mint_url() {
            Ok((None, local_keysets_info))
```

stefanbitcr (approved):

PR #197, opened 2026-04-29
Use new Coin Selection and Fees

zupzup:
Implements:
* Coin Selection from bcr-common
* Receiving and Sending of mixed (credit and debit keysets) tokens
* Swaps with fees (based on the current impl in bcr-common)
* Small reliability improvements for detecting rabid / migrating from rabid

### DIFF
--- a/crates/bcr-wallet-api/src/pocket/debit.rs
+++ b/crates/bcr-wallet-api/src/pocket/debit.rs
@@ -10,6 +10,7 @@ use bcr_common::{
         self, Amount, CurrencyUnit, KeySet, KeySetInfo, Proof, ProofsMethods, amount::SplitTarget,
         nut00 as cdk00, nut01 as cdk01,
     },
+    core::swap::wallet::{PaymentPlan, prepare_payment},
     wire::{common as wire_common, melt as wire_melt, mint as wire_mint, swap as wire_swap},
 };
 use bcr_wallet_core::types::{MeltSummary, MintSummary, Seed, SendSummary};
@@ -141,30 +142,37 @@ impl Pocket {
         }
     }
 
-    fn find_active_keysetid(&self, keysets_info: &[KeySetInfo]) -> Result<cashu::KeySetInfo> {
+    fn validate_keysets<'inf>(
+        &self,
+        keysets_info: &'inf [KeySetInfo],
+        inputs: &[cdk00::Proof],
+    ) -> Result<HashMap<cashu::Id, &'inf KeySetInfo>> {
+        let infos = collect_keyset_infos_from_proofs(inputs.iter(), keysets_info)?;
+        for info in infos.values() {
+            if info.unit != self.unit {
+                return Err(Error::InvalidCurrencyUnit(info.unit.clone().to_string()));
+            }
+            if !info.active {
+                return Err(Error::InactiveKeyset(info.id));
+            }
+        }
+        Ok(infos)
+    }
+
+    fn find_debit_keysetid(&self, keysets_info: &[KeySetInfo]) -> Result<cashu::KeySetInfo> {
         let active_info = keysets_info
             .iter()
-            .find(|info| info.unit == self.unit && info.active);
+            .find(|info| info.unit == self.unit && info.active && info.final_expiry.is_none());
         let Some(active_info) = active_info else {
             return Err(Error::NoActiveKeyset);
         };
         Ok(active_info.clone())
     }
 
-    async fn find_active_keyset(
-        &self,
-        keysets_info: &[KeySetInfo],
-        client: &Arc<dyn ClowderMintConnector>,
-    ) -> Result<(KeySetInfo, KeySet)> {
-        let active_info = self.find_active_keysetid(keysets_info)?;
-        let active_keyset = client.get_mint_keyset(active_info.id).await?;
-        Ok((active_info, active_keyset))
-    }
-
     async fn digest_proofs(
         &self,
         client: Arc<dyn ClowderMintConnector>,
-        (active_info, active_keyset): (cashu::KeySetInfo, cashu::KeySet),
+        keysets_info: &[KeySetInfo],
         inputs: HashMap<cdk01::PublicKey, cdk00::Proof>,
         swap_config: SwapConfig,
     ) -> Result<(Amount, Vec<cdk01::PublicKey>)> {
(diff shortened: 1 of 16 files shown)


mtbitcr (approved):

crates/bcr-wallet-api/src/pocket/debit.rs
```diff
@@ -141,55 +142,100 @@ impl Pocket {
...
-            .increment_counter(active_info.id, counter, premint_secrets.len() as u32)
-            .await?;
-        let premints = HashMap::from([(active_info.id, premint_secrets)]);
-        let keysets = HashMap::from([(active_info.id, active_keyset)]);
+        // prepare data
+        let kinfos: HashMap<cashu::Id, KeySetInfo> =
+            keysets_info.iter().cloned().map(|k| (k.id, k)).collect();
+        let ys = inputs.keys().cloned().collect();
+        let mut old_proofs: HashMap<cashu::Id, Vec<cdk00::Proof>> = HashMap::new();
+        let mut swap_proofs: Vec<cdk00::Proof> = Vec::new();
```
codingpeanut157: it looks like you can optimize this with a 
```rust
let (ys, swap_proofs) : ( Vec<_>, Vec<_>) = inputs.into_iter().unzip();
```

crates/bcr-wallet-api/src/pocket/debit.rs
```diff
@@ -141,55 +142,100 @@ impl Pocket {
...
+        }
+
+        // create swap plan
+        let swap_plan = prepare_swap(&swap_proofs, &kinfos)?;
+        tracing::debug!("Digest proofs - swap plan: {swap_plan:?}");
+
+        // collect keysets first as we don't want any failure once the swap request
+        // has been made
+        let mut keysets: HashMap<cashu::Id, KeySet> = HashMap::new();
+        for kid in old_proofs.keys() {
```
codingpeanut157: if you use `old_proofs` only for its keys, why not collecting unique kids with
```rust
let kids: HashSet<cashu::Id> = swap_proofs.iter().map(|p| p.keyset_id.clone()).collect();
```

crates/bcr-wallet-api/src/pocket/debit.rs
```diff
@@ -251,8 +316,7 @@ impl Pocket {
         client: Arc<dyn ClowderMintConnector>,
         swap_config: SwapConfig,
     ) -> Result<(cashu::Amount, Vec<cashu::PublicKey>)> {
-        let (active_keyset_info, active_keyset) =
-            self.find_active_keyset(keysets_info, &client).await?;
+        let (_, active_keyset) = self.find_debit_keyset(keysets_info, &client).await?;
```
codingpeanut157: you don't need to call this
`PremintSecrets` already has the keyset_id field you can use to retrieve the keyset through the client

codingpeanut157 (approved):
A few notes:
#### Proofs
Proofs might be heavy in terms of their footprints, we should try to minimize the cloning

#### KeysetInfo
I can see there is a bit of friction with the `&[KeysetInfo]`
callers pass a `&[KeysetInfo]` while some callees want a `&HashMap<cashu::Id, KeysetInfo>`
Maybe we should fix that and move to the most sophisticated option between the two, which is the hashmap, so you don't have to filter and clone, and you can just pass around references to the same hashmap.
At some point we might also think about optimizing the local caching of this.
A possible idea is to have a mint endpoint that returns the hash of a sorted `Vec<cashu::Id>` so wallets can easily check if their local cache is updated. Feel free to explore the idea and open tickets accordingly

#### Debit keyset
I am going to write a couple of functions in bcr-commn that will do proof-selection for melting, and keysetID selection for minting, so the wallet does not have to worry about `find_debit_keyset`.
The policies for such selection might change, so I don't want the wallet to worry that much about.

crates/bcr-wallet-api/src/pocket/debit.rs
```diff
@@ -141,55 +142,100 @@ impl Pocket {
...
+        // create swap plan
+        let swap_plan = prepare_swap(&swap_proofs, &kinfos)?;
+        tracing::debug!("Digest proofs - swap plan: {swap_plan:?}");
+
+        // collect keysets first as we don't want any failure once the swap request
+        // has been made
+        let mut keysets: HashMap<cashu::Id, KeySet> = HashMap::new();
+        for kid in old_proofs.keys() {
+            let keyset = client.get_mint_keyset(*kid).await?;
+            keysets.insert(*kid, keyset);
```
stefanbitcr, line 198: Swapping to the same keyset is the correct behavior, what do we do though if we do need to deactivate a keyset ?

crates/bcr-wallet-api/src/pocket/debit.rs
```diff
@@ -141,55 +142,100 @@ impl Pocket {
...
+        &self,
+        keysets_info: &'inf [KeySetInfo],
+        inputs: &[cdk00::Proof],
+    ) -> Result<HashMap<cashu::Id, &'inf KeySetInfo>> {
+        let infos = collect_keyset_infos_from_proofs(inputs.iter(), keysets_info)?;
+        for info in infos.values() {
+            if info.unit != self.unit {
+                return Err(Error::InvalidCurrencyUnit(info.unit.clone().to_string()));
+            }
+            if !info.active {
```
stefanbitcr, line 155: Do we reject inactive keysets when receiving ? Should we allow inactive -> active swaps ?
Question also for @codingpeanut157

stefanbitcr (approved):

crates/bcr-wallet-api/src/pocket/debit.rs
codingpeanut157: I think for the time being we decided not to deactivate any keyset
In the future we will figure out how to do it.

My plan is to move all the logic related to keysets in the bcr-common::core library so the wallet does not have to worry about it that much

crates/bcr-wallet-api/src/pocket/debit.rs
codingpeanut157: no key should be inactive for now.
But yeah, we could add some logic in the validating functions

zupzup:
> A few notes:
> #### Proofs
> 
> Proofs might be heavy in terms of their footprints, we should try to minimize the cloning
> #### KeysetInfo
> 
> I can see there is a bit of friction with the `&[KeysetInfo]` callers pass a `&[KeysetInfo]` while some callees want a `&HashMap<cashu::Id, KeysetInfo>` Maybe we should fix that and move to the most sophisticated option between the two, which is the hashmap, so you don't have to filter and clone, and you can just pass around references to the same hashmap. At some point we might also think about optimizing the local caching of this. A possible idea is to have a mint endpoint that returns the hash of a sorted `Vec<KeysetInfo>` so wallets can easily check if their local cache is updated. Feel free to explore the idea and open tickets accordingly
> #### Debit keyset
> 
> I am going to write a couple of functions in bcr-commn that will do proof-selection for melting, and keysetID selection for minting, so the wallet does not have to worry about `find_debit_keyset`. The policies for such selection might change, so I don't want the wallet to worry that much about.

Very good input, I created an improvement task to streamline keysets and caching as well as proof cloning. :+1:

codingpeanut157 (approved):

PR #205, opened 2026-05-06
remove any trace of cdk crate

codingpeanut157:
replace the source of some types from cdk to cdk-common

We are about to remove bcr-common::cdk


zupzup (approved):
👍

stefanbitcr (approved):

mtbitcr (approved):

PR #206, opened 2026-05-06
Add support for multiple Wallets

zupzup:
* Add support for multiple wallets
    * `default_mint_url`, `bitcoin_network`, `mnemonic`, `nostr_relays` are now wallet-specific, not application-specific
    * `WalletFfiConfig` now takes a `HashMap<WalletId, Mnemonic>` where existing wallets need to be added with their ids and `mnemonics`
    * Wallet functions are not called with a wallet index anymore, but with the `wallet_id`
    * The `wallet_id` is the hashed seed (of the `mnemonic`) + `bitcoin_network`
        * This means a combination of `mnemonic` + `bitcoin_network` is unique
    * To get the `wallet_id` and a random `mnemonic` before creating a wallet on a given `bitcoin_network`, you can use the `generate_random_mnemonic` call, which now returns the `wallet_id` and the `mnemonic`
    * To get the `wallet_id` for a given `mnemonic` and `bitcoin_network`, you can call `wallet_id_for_mnemonic_and_network`
    * The Wallet ID must be unique per application
    * The Wallet Name must be unique per application
    * `wallet_delete` now properly deletes all tables of the given wallet
    * Nostr clients are now running per-wallet, instead of one per application
* Add proper error logging to external mint calls
* Upgrade dependencies


codingpeanut157 (approved):
Are we sure we want a different mnemonic for every wallet?
Wouldn't be better to derive all the wallets from the same mnemonic?

zupzup:
> Are we sure we want a different mnemonic for every wallet? Wouldn't be better to derive all the wallets from the same mnemonic?

The current idea is to have maximal flexibility in terms of btc network, mnemonics, mints etc., hence these requirements. (cc @mtbitcr)

I guess there's a point to be made for either approach.

* One mnemonic is nice for backup/restore (only one thing to remember/save)
* Multiple mnemonics enable more flexibility in case someone simply has (or wants to have) multiple wallets with different mnemonics (e.g. one private, one company etc.)

I guess it would also be possible to create a backup mechanism that enables the first advantage (one mnemonic for the user, wallet-mnemonics are encrypted and backed up with this one mnemonic via Nostr etc.)

Issue #164, opened 2026-02-10
Improve Performance (especially HTTP)


zupzup:
* When we do HTTP requests to multiple betas, they should be in parallel
* We should make as few HTTP requests as possible
  * Fetch data on-demand
  * Cache data
  * => This is also important for Offline Mode
* Create indices for database tables based on querying patterns (mostly relevant for transactions and proofs)
* MAYBE (experiment if it makes sense): Use `block_in_place` instead of `spawn_blocking` for redb (https://docs.rs/tokio/latest/tokio/task/fn.block_in_place.html / https://github.com/cberner/redb/issues/30)

PR #220, opened 2026-05-25
API for Filtering and Paging Transactions

zupzup:
* Adapt `wallet_get_transactions` to be filtered and paged - takes filters, sorting and a limit and can be progressed using a returned cursor
    * Can be filtered by `payment_type`, `status`, `direction` and a range for `timestamp`
    * Can be sorted by `timestamp` and `amount` ascending and descending
    * There is a `limit`, to set the amount of items returned, clamped between 10 - 200 per call
    * A `TransactionCursor` is returned per call, which can be passed into the next request to get the next page
        * if the `next_cursor` is empty, all results have been returned


crates/bcr-wallet-core/src/types.rs
```diff
@@ -155,3 +155,124 @@ impl std::convert::From<MeltSummary> for PaymentSummary {
...
+#[derive(Debug, Clone, Default)]
+pub struct TransactionFilters {
+    pub payment_types: Vec<PaymentType>,
+    pub statuses: Vec<TransactionStatus>,
+    pub direction: Option<TransactionDirection>,
+    pub time_range: Option<TimeRange>,
+}
+
+impl TransactionFilters {
+    pub fn matches_tx(&self, tx: &Transaction) -> bool {
```
codingpeanut157, line 168: uhm... this might become slow if the number of transactions grow
For each non-filtering value you are "wasting" an `if let` for each transaction

Maybe it's not an issue given the volume of transactions.
But videogame-programming would suggest another, less elegant but more effective approach

```rust
    fn matches_tx(&self, txs: &mut Vec<Transaction>) {
        if let Some(fa) = self.fa {
            txs.retain(|o| o.a < fa)
        }
        if let Some(fb) = self.fb {
            txs.retain(|o| o.b < fb)
        }
    }
```
`calling` retain multiple times it's not ideal either ( a lot of Transactions moved around), yet, you are paying a constant time for any non-existing filter.
And if the filtering element is relevant ( let's say it cuts the amount of transactions in 1/2 ) the retains is getting called on a logn-decreasing amount of elements

The even more effective, would be to manually explore all the filtering options and build a closure that filters based on the filter combination... maybe too much

crates/bcr-wallet-core/src/types.rs
```diff
@@ -155,3 +155,124 @@ impl std::convert::From<MeltSummary> for PaymentSummary {
...
+pub enum TransactionSort {
+    TimeAsc,
+    #[default]
+    TimeDesc,
+    AmountAsc,
+    AmountDesc,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum TransactionCursor {
```
codingpeanut157, line 214: you can use `strum` here and do
```rust
#[derive(Debug, Clone, PartialEq, Eq, strum::EnumDiscriminants)]
#[strum_discriminants(name(TransactionSort))]
pub enum TransactionCursor {
    TimeDesc { tstamp: u64, id: TransactionId },
    TimeAsc { tstamp: u64, id: TransactionId },
    AmountDesc { amount: Amount, id: TransactionId },
    AmountAsc { amount: Amount, id: TransactionId },
}
```
and you get a `from`  function from `TransactionCursor` to `TransactionSort` for free

codingpeanut157 (approved):

crates/bcr-wallet-core/src/types.rs
zupzup: That would certainly be more cache-friendly, yes. But I'm not sure it matches the usage-pattern, which will be e.g. fetching ~25 (a phone page full) of entries at a time.

The actual performance bottleneck is that we fetch all transactions from the DB on each call, so using some indices there based on the filtering patterns would be the biggest win by far, heavily reducing the amount of transactions to filter.

This is basically just a naive first implementation that saves the client from getting all transactions on each call to implement paging/infinite scrolling.

zupzup: This future optimization (indices etc.) is already part of the general Wallet-Performance-Optimization task :D 
https://github.com/BitcreditProtocol/Wallet-Core/issues/164

crates/bcr-wallet-core/src/types.rs
```diff
@@ -155,3 +155,125 @@ impl std::convert::From<MeltSummary> for PaymentSummary {
...
+                id: tx.id(),
+            },
+            TransactionSort::AmountDesc => Self::AmountDesc {
+                amount: tx.amount,
+                id: tx.id(),
+            },
+        }
+    }
+
+    // check if the given transaction is after the given cursor, falling back to comparing transaction ids
```
stefanbitcr, line 262: Wouldn't a better fallback be (amount, timestamp, id):
* Better DB index than random hashes,

crates/bcr-wallet-ffi/src/api/mod.rs
```diff
@@ -639,12 +639,22 @@ pub async fn wallet_get_transaction_ids(
...
 pub async fn wallet_get_transactions(
-    req: WalletRequest,
-) -> Result<WalletTransactionsResponse, WalletError> {
+    req: WalletListTransactionsRequest,
+) -> Result<WalletListTransactionsResponse, WalletError> {
     let app_state = get_app_state().await;
-    let ids = app_state.wallet_list_txs(req.wallet_id).await?;
-    Ok(WalletTransactionsResponse {
-        txs: ids.into_iter().map(|t| t.into()).collect(),
+    let limit = req.limit.clamp(10, 200);
```
stefanbitcr: Is limit validation only on FFI layer ?

stefanbitcr (approved):

crates/bcr-wallet-ffi/src/api/mod.rs
zupzup: yeah, I didn't want to impose this on pure library users, but I guess for the wallet, 5-100 would actually be a more suitable range anyway (phone screens etc.)

crates/bcr-wallet-core/src/types.rs
zupzup: They are distinct cursor cases and this gives the secondary global id sorting. But yeah, once this is moved to use (likely several) indices, this will change a bit anyway

PR #222, opened 2026-05-26
Add fees overview by month to transaction list

zupzup:
* Add a `fees_by_month` field to the result of `wallet_get_transactions`, which sums up the fees for the distinct months present in the returned transactions by UTC time, sorted descending

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
     * There is a `limit`, to set the amount of items returned, clamped between 5 - 100 per call
     * A `TransactionCursor` is returned per call, which can be passed into the next request to get the next page
         * if the `next_cursor` is empty, all results have been returned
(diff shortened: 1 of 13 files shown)


codingpeanut157 (approved):
honestly, it looks a bit weird...

What if I filter with a time range of 15 days? FeesByMonth loses meaning, no?
It looks like this was more of a frontend requests that instead of being implemented in the frontend leaked into backend logic.

zupzup:
> honestly, it looks a bit weird...
> 
> What if I filter with a time range of 15 days? FeesByMonth loses meaning, no? It looks like this was more of a frontend requests that instead of being implemented in the frontend leaked into backend logic.

Yeah that's a fair analysis :sweat_smile:, the initial idea was to have an overview of fees per month for the user, but with filters and unfinished months, it's not clear yet if it makes sense to show it and generally, if it doesn't make more sense to calculate it on the frontend depending on the context.

But for a first approximation, this should be fine and I have no qualms about ripping it out again, should it prove to not be useful. :v:

PR #224, opened 2026-05-28
Introduce Transport Crate for Nostr

zupzup:
First step of introducing a proper Nostr setup

* Create a `bcr-wallet-transport` crate that will contain all nostr-specific logic
* Creata a `NostrClient` abstraction and move existing logic there

Next steps:

* persistence for nostr events inbox/outbox
* proper metadata publishing and handling like in E-Bill

All in preparation of implementing pay-to-nostr-contact

### DIFF
--- a/crates/bcr-wallet-transport/src/lib.rs
+++ b/crates/bcr-wallet-transport/src/lib.rs
@@ -1,0 +1,174 @@
+use std::{collections::HashMap, sync::Arc};
+
+use bcr_common::cashu::{Amount, ProofsMethods, nut18 as cdk18};
+use bcr_wallet_core::types::{
+    PAYMENT_TYPE_METADATA_KEY, PaymentType, TRANSACTION_STATUS_METADATA_KEY, TransactionStatus,
+};
+use error::Result;
+use nostr::{
+    event::EventId,
+    nips::{
+        nip19::{FromBech32, Nip19Profile, ToBech32},
+        nip59::UnwrappedGift,
+    },
+    signer::NostrSigner,
+    types::RelayUrl,
+};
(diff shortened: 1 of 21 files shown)


crates/bcr-wallet-transport/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    pub async fn handle_event(
+        &self,
+        received_evt: RelayPoolNotification,
+        payment_id: Uuid,
+        expected: Amount,
+    ) -> Result<Option<(cdk18::PaymentRequestPayload, HashMap<String, String>)>> {
+        let signer = self.signer().await?;
+        if let RelayPoolNotification::Event { event, .. } = received_evt {
```
codingpeanut157: may I suggest
```rust
let RelayPoolNotification::Event { event, .. } = received_evt  else {
    return Ok(None);
}
if event.kind ...
```

it saves one extra indentation for the rest of the function and shorten the alternative path to the `else`

crates/bcr-wallet-transport/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+            if event.kind != nostr_sdk::Kind::GiftWrap {
+                tracing::debug!("handle event, but no GiftWrap - {}", event.kind);
+                return Ok(None);
+            }
+
+            let payload = match UnwrappedGift::from_gift_wrap(&signer, &event).await {
```
codingpeanut157: same here
```rust
let UnwrappedGift { rumor, .. } = match UnwrappedGift::from_gift_wrap(&signer, &event).await {
    Ok(gift) => gift,
    Err(e) => {
        tracing::debug! (e);
        return Ok(None);
    };
};
if rumor ...
```

codingpeanut157 (approved):

PR #231, opened 2026-06-02
Nostr Event Storage & Nostr Consumer

zupzup:
* Add proper `NostrClient` and `NostrConsumer` abstractions with useful utilities (heavily ~copied~ inspired by E-Bill)
* Move event handling to `NostrConsumer`
  * Communication between the app and the nostr consumer happens via a channel
* Add a mechanism to avoid handling events multiple times (event offset)
* Fix nostr private key derivation differing from E-Bill

Next steps:

* Expose `connected` state to frontend
* Retry Queue
* Proper Relay list synchronization as to not conflict with E-Bill there
* Contacts etc.

### DIFF
--- a/crates/bcr-wallet-persistence/src/redb/nostr.rs
+++ b/crates/bcr-wallet-persistence/src/redb/nostr.rs
@@ -1,0 +1,348 @@
+use crate::{NostrEventOffset, NostrEventOffsetStoreApi, error::Result};
+use async_trait::async_trait;
+use nostr::types::Timestamp;
+use redb::{Database, ReadableDatabase, ReadableTable, TableDefinition, TableError};
+use std::sync::Arc;
+use tokio::task::spawn_blocking;
+
+///////////////////////////////////////////// NostrEventOffsetEntry
+#[derive(serde::Serialize, serde::Deserialize, Debug, Clone)]
+pub struct NostrEventOffsetEntry {
+    pub event_id: String,
+    pub time: Timestamp,
+    pub success: bool,
+}
+
+impl From<NostrEventOffsetEntry> for NostrEventOffset {
+    fn from(value: NostrEventOffsetEntry) -> Self {
+        Self {
+            event_id: value.event_id,
+            time: value.time,
+            success: value.success,
+        }
+    }
+}
+
+impl From<NostrEventOffset> for NostrEventOffsetEntry {
+    fn from(value: NostrEventOffset) -> Self {
+        Self {
+            event_id: value.event_id,
+            time: value.time,
+            success: value.success,
+        }
+    }
+}
+
+///////////////////////////////////////////// NostrDB
+pub struct NostrDB {
+    db: Arc<Database>,
+    last_offset_table: TableDefinition<'static, &'static [u8], u64>,
+    offset_entry_table: TableDefinition<'static, &'static [u8], Vec<u8>>,
+}
+
+impl NostrDB {
+    const OFFSET_ENTRY_DB_NAME: &'static str = "offset_entry";
+    const LAST_OFFSET_DB_NAME: &'static str = "last_offset";
+    const CURRENT_OFFSET_UNIQUE_ID: &'static str = "current_offset";
+
+    pub fn new(db: Arc<Database>, wallet_id: &str) -> Result<Self> {
+        // Leak once to get static string, because of dynamically generated table names
+        let offset_entry_name: &'static str =
+            Box::leak(format!("{wallet_id}_{}", Self::OFFSET_ENTRY_DB_NAME).into_boxed_str());
+        // Leak once to get static string, because of dynamically generated table names
+        let last_offset_name: &'static str =
+            Box::leak(format!("{wallet_id}_{}", Self::LAST_OFFSET_DB_NAME).into_boxed_str());
+
+        let offset_entry_table = TableDefinition::new(offset_entry_name);
+        let last_offset_table = TableDefinition::new(last_offset_name);
+
+        Ok(Self {
+            db,
+            offset_entry_table,
+            last_offset_table,
+        })
+    }
+
+    fn current_offset_sync(
+        db: Arc<Database>,
+        last_offset_table: TableDefinition<'static, &'static [u8], u64>,
+    ) -> Result<Timestamp> {
+        let read_txn = db.begin_read()?;
+
+        match read_txn.open_table(last_offset_table) {
+            Ok(table) => {
+                let entry = table.get(Self::CURRENT_OFFSET_UNIQUE_ID.as_bytes())?;
+                match entry {
+                    Some(t) => Ok(Timestamp::from(t.value())),
+                    None => Ok(Timestamp::zero()),
+                }
+            }
+            Err(TableError::TableDoesNotExist(_)) => Ok(Timestamp::zero()),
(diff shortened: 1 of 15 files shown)


mtbitcr (approved):

crates/bcr-wallet-transport/src/nostr.rs
```diff
@@ -0,0 +1,390 @@
...
+    sync::{
+        Arc,
+        atomic::{AtomicBool, Ordering},
+    },
+    time::Duration,
+};
+use tokio::task::JoinSet;
+
+#[derive(Clone)]
+pub struct NostrClient {
```
codingpeanut157: nitpicking
the `NostrClient` it's already in the `nostr` module
`nostr::NostrClient` sounds a bit redundant

crates/bcr-wallet-transport/src/nostr.rs
```diff
@@ -0,0 +1,390 @@
...
+    client: Client,
+    relays: Vec<RelayUrl>,
```
codingpeanut157, line 34: nitpicking

if the set of relays never changes, an `relays: Arc<[RelayUrl]>` is much more efficient `clone()`-wise

crates/bcr-wallet-transport/src/nostr.rs
```diff
@@ -0,0 +1,390 @@
...
+        &self,
+        filter: Filter,
+        order: Option<SortOrder>,
+        relays: Option<Vec<RelayUrl>>,
+    ) -> Result<Vec<Event>> {
+        let events = self
+            .client()
+            .await?
+            .fetch_events_from(
+                relays.unwrap_or(self.relays.clone()),
```
codingpeanut157, line 152: if you use `Arc<[T]>` as suggested above, here you could do
```rust
let cloned_default_relays = self.relays.clone();
unwrap_or_else(move|| {cloned_default_relays.to_vec()})
```
and save a potential extra clone

crates/bcr-wallet-transport/src/nostr.rs
```diff
@@ -0,0 +1,390 @@
...
+    async fn client(&self) -> Result<&Client> {
+        if !self.is_connected() {
+            self.connect().await?;
+        }
+        Ok(&self.client)
+    }
+}
+
+#[derive(Clone)]
+pub struct NostrConsumer {
```
codingpeanut157: nitpicking
``nostr::NostrConsumer` seems a bit repetitive

crates/bcr-wallet-transport/src/nostr.rs
```diff
@@ -0,0 +1,390 @@
...
+            .pubkey(client.signer.public_key());
+
+        client.subscribe(nostr_filter).await.map_err(|e| {
+            tracing::error!("Failed to subscribe to Nostr public events: {e}");
+            Error::Network("Failed to subscribe to Nostr public events".to_string())
+        })?;
+
+        let client_for_task = client.clone();
+        let offset_store_clone = self.offset_store.clone();
+        let signer = self.client.signer().await?;
```
codingpeanut157, line 259: why the `let client = self.client.clone();` on the first line of this function then?

crates/bcr-wallet-transport/src/nostr.rs
```diff
@@ -0,0 +1,390 @@
...
+        let event_channel = self.event_channel.clone();
+        tasks.spawn(async move {
+            client_for_task
+                .client
+                .handle_notifications(move |note| {
+                    let offset_store = offset_store_clone.clone();
```
codingpeanut157, line 265: why the clone of a clone? is the outer closure called multiple times by `handle_notifications`?

crates/bcr-wallet-transport/src/nostr.rs
```diff
@@ -0,0 +1,390 @@
...
+    })
+    .await
+    .map_err(|e| tracing::error!("Could not store event offset: {e}"))
+    .ok();
+}
+
+async fn get_offset(db: &Arc<dyn NostrEventOffsetStoreApi>) -> Timestamp {
+    db.current_offset()
+        .await
+        .map_err(|e| tracing::error!("Could not get event offset: {e}"))
```
codingpeanut157: maybe you can do
```rust
unwrap_or_else (|e| {
tracing::error
Timestamp::zero()
})

File: crates/bcr-wallet-transport/src/nostr.rs [hunk 8]
```diff
@@ -0,0 +1,390 @@
...
+) -> Result<(bool, Timestamp)> {
+    tracing::debug!("Processing Nostr event with id: {}", event.id);
+    if event.kind != nostr_sdk::Kind::GiftWrap {
+        tracing::debug!("handle event, but no GiftWrap - {}", event.kind);
+        return Ok((false, Timestamp::zero()));
+    }
+
+    let UnwrappedGift { rumor, .. } = match UnwrappedGift::from_gift_wrap(&signer, &event).await {
+        Ok(gift) => gift,
+        Err(e) => {
```
codingpeanut157, line 334: isn't this error also reporting if the event is not a GiiftWrap?
can we avoid the first if and just rely on this `match`?

crates/bcr-wallet-transport/src/lib.rs
```diff
@@ -1,174 +1,81 @@
...
-                receiver.public_key,
-                payload,
-                std::iter::empty(),
-            )
-            .await?;
-        Ok(output.id().to_owned())
-    }
+#[async_trait]
+pub trait TransportClientApi: SendSync {
+    async fn client(&self) -> Result<&Client>;
```
codingpeanut157: 🫣
would it be possible to avoid exposing the internals like the real nostr::Client?

crates/bcr-wallet-persistence/src/redb/nostr.rs
```diff
@@ -0,0 +1,348 @@
...
+
+impl NostrDB {
+    const OFFSET_ENTRY_DB_NAME: &'static str = "offset_entry";
+    const LAST_OFFSET_DB_NAME: &'static str = "last_offset";
+    const CURRENT_OFFSET_UNIQUE_ID: &'static str = "current_offset";
+
+    pub fn new(db: Arc<Database>, wallet_id: &str) -> Result<Self> {
+        // Leak once to get static string, because of dynamically generated table names
+        let offset_entry_name: &'static str =
+            Box::leak(format!("{wallet_id}_{}", Self::OFFSET_ENTRY_DB_NAME).into_boxed_str());
```
codingpeanut157, line 51: `String` has `leak` too
[ref](https://doc.rust-lang.org/std/string/struct.String.html#method.leak)

codingpeanut157 (approved):

crates/bcr-wallet-transport/src/nostr.rs
zupzup: They will be able to change, though (manually and automatically, if they are set differently from another client)

crates/bcr-wallet-transport/src/nostr.rs
zupzup: `handle_notifications` takes an `Fn`, not an `FnOnce`, so there's no guarantee that it isn't called multiple times, which makes sense, since the closure is called for each notification individually

PR #233, opened 2026-06-04
Nostr Retry Queue and other improvements

zupzup:
* Structure Nostr as Consumer and Transport
* Add a retry-mechanism
* Expose nostr connection status to frontend
* Synchronize with published relay-list
* Add a background-receiver for incoming nostr payments

### DIFF
--- a/crates/bcr-wallet-api/src/wallet/mod.rs
+++ b/crates/bcr-wallet-api/src/wallet/mod.rs
@@ -32,10 +32,8 @@ use bitcoin::{
     secp256k1,
 };
 use std::{collections::HashMap, str::FromStr, sync::Arc, time::Duration};
-use tokio::{
-    sync::{Mutex, broadcast},
-    task::JoinHandle,
-};
+use tokio::sync::{Mutex, broadcast};
+use tokio_util::sync::CancellationToken;
 
 pub struct Wallet {
     network: bitcoin::Network,
@@ -56,7 +54,7 @@ pub struct Wallet {
     nostr_event_channel: NostrEventChannel,
     nostr_repo: Arc<dyn NostrRepository>,
     nostr_consumer_running: Arc<Mutex<bool>>,
-    nostr_handle: Option<JoinHandle<()>>,
+    nostr_shutdown: CancellationToken,
 }
 
 impl Wallet {
@@ -78,7 +76,8 @@ impl Wallet {
         nostr_repo: Arc<dyn NostrRepository>,
         nostr_consumer: nostr::Consumer,
     ) -> Self {
-        let mut wallet = Self {
+        let cancel = CancellationToken::new();
+        let wallet = Self {
             network,
             client,
             mint_keyset_infos,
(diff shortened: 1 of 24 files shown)


crates/bcr-wallet-api/src/wallet/mod.rs
```diff
@@ -92,12 +93,76 @@ impl Wallet {
...
+        wallet.start_nostr_event_listener().await;
+        wallet
+    }
+
+    async fn nostr_connect(&mut self, nostr_consumer: nostr::Consumer) {
+        let wallet_id = self.id.clone();
+        let nostr_consumer_running = self.nostr_consumer_running.clone();
+        let nostr_handle = tokio::spawn(async move {
+            *nostr_consumer_running.lock().await = false;
+            loop {
```
codingpeanut157, line 112: I don't see any break or return, how is this loop supposed to be terminated?

Also, do we want to trace with this granularity `nostr_consumer_running` ?
I see you set it to true AFTER start and back to false before waiting for 10 seconds.

crates/bcr-wallet-persistence/src/redb/nostr.rs
```diff
@@ -137,10 +185,122 @@ impl NostrDB {
         Ok(())
     }
 
-    fn delete_repo(
+    ///////////////////////////////////////////// Event Offset
+    fn add_retry_message_sync(
```
codingpeanut157, line 196: `retry_message` is a bit misleading
I think [dlq](https://en.wikipedia.org/wiki/Dead_letter_queue) is what you are looking for

codingpeanut157 (approved):

crates/bcr-wallet-api/src/wallet/mod.rs
zupzup: The `handle.join_next()` call blocks on the consumer returning (e.g. if it fails somehow)

This is basically just a long-running background-reconnect loop for the consumer.

The only way it really gets dropped is when the wallet gets deleted, or on app shutdown.

`nostr_consumer_running` is just for detecting if the consumer is running from the outside, so if it's "listening". Hence it should only be true when that is given.

For context, this whole thing should practically almost never happen, and the 10 seconds is to ensure we don't try to reconnect very fast if it ever happens, but back off a bit before we do.

crates/bcr-wallet-persistence/src/redb/nostr.rs
zupzup: I thought it's only a dead letter if it couldn't be delivered (for any reason) :thinking: so basically in this case, when it was retried > max_retries.

I just delete it in that case, but it might be prudent to store those events in a separate dead_letter_table :thinking:

crates/bcr-wallet-api/src/wallet/mod.rs
zupzup: But now you mention it, it would definitely be cleaner to have a shutdown signal in there, instead of this handle + abort business. Will change :+1:

PR #234, opened 2026-06-05
Contact CRUD + List

zupzup:
* Add Contact CRUD API
    * Contacts have a `node_id` and a `name`
    * `add_contact`
    * `edit_contact`
    * `delete_contact`
    * `get_contact`
    * `list_contacts` with an optional, case-insensitive search-term
* Add tests for nostr retry queue

### DIFF
--- a/crates/bcr-wallet-api/src/wallet/api.rs
+++ b/crates/bcr-wallet-api/src/wallet/api.rs
@@ -15,11 +15,14 @@ use async_trait::async_trait;
 use bcr_common::{
     cashu::{self, Amount, CurrencyUnit, ProofsMethods, nut00 as cdk00, nut18 as cdk18},
     cdk_common::wallet::{Transaction, TransactionDirection, TransactionId},
+    core::NodeId,
     wallet::Token,
     wire::clowder::{self as wire_clowder},
 };
 use bcr_wallet_core::{
     SendSync,
+    contact::Contact,
+    name::Name,
     types::{BTC_TX_ID_TYPE_METADATA_KEY, PaymentResultCallback, PaymentType, TransactionStatus},
     util::{from_mint_url, to_mint_url},
 };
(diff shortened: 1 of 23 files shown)


crates/bcr-wallet-api/src/wallet/api.rs
```diff
@@ -1106,4 +1115,49 @@ impl WalletApi for super::Wallet {
 
         Ok(())
     }
+
+    async fn add_contact(&self, node_id: NodeId, name: Name) -> Result<()> {
+        let contact = Contact { node_id, name };
+        self.contact_repo.store_contact(contact).await?;
```
codingpeanut157: is it ok to overwrite an existing node_id, or does the db fail?

crates/bcr-wallet-api/src/wallet/api.rs
```diff
@@ -1106,4 +1115,49 @@ impl WalletApi for super::Wallet {
...
+    }
+
+    async fn delete_contact(&self, node_id: NodeId) -> Result<()> {
+        if self
+            .contact_repo
+            .get_contact(node_id.clone())
+            .await?
+            .is_some()
+        {
+            self.contact_repo.delete_contact(node_id).await?;
```
codingpeanut157, line 1154: 🤔 

what if we silently carry on if the contact does not exist?
We can spare ourselves from dealing with an error that does not really give much meaningful information to the UI

crates/bcr-wallet-core/src/name.rs
```diff
@@ -0,0 +1,124 @@
...
+
+impl FromStr for Name {
+    type Err = ValidationError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Name::new(s)
+    }
+}
+
+impl TryFrom<String> for Name {
```
codingpeanut157, line 47: can you not implement `impl From<String> ` and use `Name::new()` ?

codingpeanut157 (approved):

crates/bcr-wallet-api/src/wallet/api.rs
zupzup: right now it doesn't fail, but it should - will change

crates/bcr-wallet-api/src/wallet/api.rs
zupzup: I had that, but then I changed it to be consistent with edit. The frontend can still match on the error and choose to ignore it :shrug:

crates/bcr-wallet-core/src/name.rs
zupzup: Name::new() also returns a Result. The idea is that the validation runs on each code path

PR #236, opened 2026-06-09
Pay to Contact

zupzup:
* Add functionality to send a private, encrypted payment to an existing contact via Nostr
    * `wallet_prepare_pay_to_contact` - taking a node_id, amount and description
    * `wallet_pay_to_contact` - taking a payment request id
* Add `node_id` to `WalletInfo`
* Add endpoint to get `node_id`

Next step:
* Add Tests
* Add Remote Request to Pay via Nostr
* Add checks for same network
* Add logic to keep contact's relays up2date


crates/bcr-wallet-api/src/purse/mod.rs
```diff
@@ -58,11 +58,11 @@ where
         res
     }
 
-    pub async fn add_wallet(&self, wallet: Wlt) -> Result<String> {
-        let wallet_id = wallet.id();
-        self.repo.store(wallet.config()?).await?;
+    pub async fn add_wallet(&self, wallet: Arc<RwLock<Wlt>>) -> Result<String> {
```
codingpeanut157, line 61: what's the reason behind this change?
Do you need to keep a reference to the wallet at the calling site?

crates/bcr-wallet-core/src/contact.rs
```diff
@@ -1,8 +1,10 @@
 use crate::name::Name;
 use bcr_common::core::NodeId;
+use nostr::types::RelayUrl;
 
 #[derive(Debug, Clone)]
 pub struct Contact {
     pub node_id: NodeId,
     pub name: Name,
+    pub nostr_relays: Vec<RelayUrl>,
```
codingpeanut157, line 9: I guess these relays are the ones used by the `Contact`, right?

maybe I have some lack of knowledge about nostr, but I thought that as long as you are using popular relays, you don't need to point your client to a specific relay in order to reach the destination of your message.
Are not relays in contact to each other?

crates/bcr-wallet-core/src/event.rs
```diff
@@ -0,0 +1,110 @@
...
+    DEFAULT_EVENT_VERSION.into()
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, BorshSerialize, BorshDeserialize)]
+pub enum EventType {
+    ContactPayment,
+}
+
+impl EventType {
+    pub fn all() -> Vec<EventType> {
```
codingpeanut157: another case for strum
https://docs.rs/strum/latest/strum/derive.VariantArray.html

crates/bcr-wallet-core/src/event.rs
```diff
@@ -0,0 +1,110 @@
...
+        vec![EventType::ContactPayment]
+    }
+}
+
+impl Display for EventType {
```
codingpeanut157: another case for strum
https://docs.rs/strum/latest/strum/derive.Display.html

crates/bcr-wallet-core/src/event.rs
```diff
@@ -0,0 +1,110 @@
...
+
+#[derive(BorshSerialize, BorshDeserialize, Debug, Clone)]
+pub struct EventEnvelope {
+    pub event_type: EventType,
+    pub version: String,
+    pub data: Vec<u8>,
+}
+
+#[derive(Debug, Clone, BorshSerialize, BorshDeserialize)]
+pub struct ContactPaymentPayload {
```
codingpeanut157, line 87: What if we send the `bcr_common::wallet::Token` alongside the sender?

crates/bcr-wallet-api/src/wallet/api.rs
```diff
@@ -294,57 +316,59 @@ impl WalletApi for super::Wallet {
                         },
                     };
 
-                    let NostrWalletEvent::Cdk18Payment { event_id, payload, sender } = received_evt;
-
-                    if payload.id != Some(p_id.to_string()) {
-                        tracing::debug!("handle event, payment id doesn't match");
-                        continue;
-                    }
+                    if let NostrWalletEvent::Cdk18Payment { event_id, payload, sender } = received_evt
```
codingpeanut157: could we use the `let-else` pattern and save one extra indentation?

crates/bcr-wallet-api/src/wallet/api.rs
```diff
@@ -294,57 +316,59 @@ impl WalletApi for super::Wallet {
...
+                        let response = <Self as api::WalletApi>::receive_proofs(
+                            self,
+                            payload.proofs,
+                            payload.unit,
+                            from_mint_url(&payload.mint),
+                            chrono::Utc::now().timestamp() as u64,
+                            payload.memo,
+                            meta,
+                        )
+                            .await;
```
codingpeanut157: 🤔 
no potential error from this function?
I am asking because I see you set the TRANSACTION_STATUS as `settled` already, before actually settling it, what happens if the mint_url does not exist, or proofs turn out to be already spent?

crates/bcr-wallet-api/src/lib.rs
```diff
@@ -343,6 +350,41 @@ impl AppState {
         })
     }
 
+    pub async fn wallet_prepare_pay_to_contact(
+        &self,
+        id: String,
```
codingpeanut157: names are a bit confusing, especially considering that everything is a string a this point
maybe `wallet_id` and `receiver_id` or `destination_id`?

codingpeanut157 (approved):
minor comments

Unrelated question: how do you feel about `cdk18::PaymentRequest` and `cdk18::Transport` 
I have the impression that they could be way more ergonomic to use.

zupzup:
> minor comments
> 
> Unrelated question: how do you feel about `cdk18::PaymentRequest` and `cdk18::Transport` I have the impression that they could be way more ergonomic to use.

We're already using cdk18 for the "synchronous" payment request via Nostr.

However, beyond abusing the `memo` or `id` field, it's not really possible to add practical metadata.

I'm also not particularly convinced by the cdk18 Transport idea. I get that it's supposed to add flexibility, but at the end of the day it just becomes a compatibility matrix again.

And since cashu interoperability isn't a goal anymore, I'm not sure sticking with nut18 even for our payment request makes so much sense vs. opting for something custom tailored to our needs.

crates/bcr-wallet-api/src/wallet/api.rs
zupzup: No, this actively waits for the payment and if it doesn't get it, or this fails, it either continues waiting, or notifies the caller.

The actual transaction is only persisted inside `receive_proofs` if everything works, so if the swap fails, or any other error happens here, no transaction is persisted. The above just sets the metadata to set in the caseof success

crates/bcr-wallet-api/src/purse/mod.rs
zupzup: I need a reference to the wallet in a background job waiting for incoming remote payments. It's not such a big change, really, since the purse already held an `Arc<RwLock<Wallet>>>`, this basically just moves that to an earlier point, so I can re-use it for the background job.

crates/bcr-wallet-core/src/contact.rs
zupzup: There is no guarantee for that, so there's no "built-in" federation between relays.

Generally, relays forward to clients, so the idea is to fetch the contact's current relays (or to keep their relay list up2date - that's a TODO for the next PR) and publish to those relays (basically the receiver's "inbox") and to your own relays (basically your "outbox" another can listen to).

codingpeanut157:
> > minor comments
> > Unrelated question: how do you feel about `cdk18::PaymentRequest` and `cdk18::Transport` I have the impression that they could be way more ergonomic to use.
> 
> We're already using cdk18 for the "synchronous" payment request via Nostr.
> 
> However, beyond abusing the `memo` or `id` field, it's not really possible to add practical metadata.
> 
> I'm also not particularly convinced by the cdk18 Transport idea. I get that it's supposed to add flexibility, but at the end of the day it just becomes a compatibility matrix again.
> 
> And since cashu interoperability isn't a goal anymore, I'm not sure sticking with nut18 even for our payment request makes so much sense vs. opting for something custom tailored to our needs.

yeah
I thought so too
We could probably design better structs for such purpose

PR #242, opened 2026-06-15
Incoming and Outgoing Payment Requests

zupzup:
* Refactor payment requests management to work for Incoming and Outgoing payment requests

This way users can manage their incoming and outgoing payment requests, watch and filter them, cancel/reject them etc.


crates/bcr-wallet-api/src/wallet/api.rs
```diff
@@ -145,25 +145,29 @@ pub trait WalletApi: SendSync {
...
+    async fn add_payment_request(
         &self,
-        pending_payment_request: PendingPaymentRequest,
+        pending_incoming_payment_request: PaymentRequest,
     ) -> Result<()>;
-    async fn prepare_pay_pending_payment_request(
+    async fn prepare_pay_payment_request(&self, payment_req_id: Uuid) -> Result<PaymentSummary>;
+    async fn reject_payment_request(&self, payment_req_id: Uuid) -> Result<()>;
+    async fn cancel_payment_request(&self, payment_req_id: Uuid) -> Result<()>;
+    async fn set_payment_request_to_paid(
```
codingpeanut157: nitpicking:
`mark_payment_request_as_paid` sounds smoother

codingpeanut157 (approved):
I guess I am falling behind with the design of the wallet APIs

As far as I understand, this PR will enable users to issue and monitor multiple payment requests (done through nostr transport, I assume)

What is the use case for this?

The only use case I can think of is remote payment requests (for online payments, assuming users nostr contacts never change)

In my head, physical point of sales and "normal" users will always process payments one at a time... Maybe there is a use case for big shops like supermarkets that have multiple cashiers, but that, I'd think, would require a whole different app with a whole different front-end and whole new protocol for interacting with the PoSs and the central wallet. 
So what is the use case for this set of new features?

Also
I don't get how a PaymentRequests can be outgoing... Am I requesting myself to pay someone? Can I not just send tokens to a contact?

zupzup:
> I guess I am falling behind with the design of the wallet APIs
> 
> As far as I understand, this PR will enable users to issue and monitor multiple payment requests (done through nostr transport, I assume)
> 
> What is the use case for this?
> 
> The only use case I can think of is remote payment requests (for online payments, assuming users nostr contacts never change)
> 
> In my head, physical point of sales and "normal" users will always process payments one at a time... Maybe there is a use case for big shops like supermarkets that have multiple cashiers, but that, I'd think, would require a whole different app with a whole different front-end and whole new protocol for interacting with the PoSs and the central wallet. So what is the use case for this set of new features?
> 
> Also I don't get how a PaymentRequests can be outgoing... Am I requesting myself to pay someone? Can I not just send tokens to a contact?

An outgoing payment request is one you send to someone else. (you expect payment)
An incoming payment request is one you receive from someone else. (you are supposed to pay)

The idea is not necessarily to have multiple "active" ones at the same time (though you could), but rather that there is a history, where you see the requests you sent and the ones you receive and whether they resulted in payments. (linked by the tx_id)

Regarding the exact use-cases, maybe @mtbitcr could shed some light on those.

codingpeanut157:
> An outgoing payment request is one you send to someone else. (you expect payment)
> An incoming payment request is one you receive from someone else. (you are supposed to pay)

That means they have the opposite meaning of payments? right?
outgoing payment: money flow out of my wallet
outgoing payment request: money ( are expected to) flow into my wallet

incoming payment: money flow into  my wallet
incoming payment request: money ( are expected to) flow out of my wallet

correct?

zupzup:
> > An outgoing payment request is one you send to someone else. (you expect payment)
> > An incoming payment request is one you receive from someone else. (you are supposed to pay)
> 
> That means they have the opposite meaning of payments? right? outgoing payment: money flow out of my wallet outgoing payment request: money ( are expected to) flow into my wallet
> 
> incoming payment: money flow into my wallet incoming payment request: money ( are expected to) flow out of my wallet
> 
> correct?

Yes :+1:

PR #246, opened 2026-06-17
Ensure Nostr Contact and Add Tests

zupzup:
* Add Tests
* Add btc network checks
* Add function to update contact relays

### DIFF
--- a/crates/bcr-wallet-api/src/wallet/mod.rs
+++ b/crates/bcr-wallet-api/src/wallet/mod.rs
@@ -1,4 +1,6 @@
 pub mod api;
+#[cfg(test)]
+pub mod test_utils;
 pub mod types;
 pub mod util;
 
@@ -16,6 +18,7 @@ use crate::{
 use bcr_common::{
     cashu::{self, Amount, CurrencyUnit, KeySetInfo, Proof, ProofsMethods},
     cdk_common::wallet::{Transaction, TransactionDirection, TransactionId},
+    core::NodeId,
     wallet::Token,
     wire::clowder::{ConnectedMintResponse, ConnectedMintsResponse},
 };
@@ -199,6 +202,7 @@ impl Wallet {
 
     async fn start_nostr_event_listener(wallet: Arc<RwLock<Self>>) {
         let mut nostr_receiver = wallet.read().await.nostr_event_channel.subscribe();
+        let wallet_network = wallet.read().await.network();
         tokio::spawn(async move {
             loop {
                 tokio::select! {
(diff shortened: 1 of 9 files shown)


crates/bcr-wallet-api/src/wallet/mod.rs
```diff
@@ -216,6 +220,10 @@ impl Wallet {
                         };
                         match received_evt {
                             bcr_wallet_transport::NostrWalletEvent::ContactPaymentRequest { sender, payload, event_id } => {
+                                if payload.sender.network() != wallet_network {
```
codingpeanut157, line 223: 🤔 

This looks like one of those silent errors that users get mad at because the bloody thing doesn't work and they can't figure out why 😅 

Do these tracing end up somewhere? Are they somehow surfaced to the users?

crates/bcr-wallet-api/src/wallet/mod.rs
```diff
@@ -1074,24 +1086,58 @@ impl Wallet {
...
     }
+
+    // check relays for a given contact and update if necessary
+    async fn ensure_nostr_contact(&self, node_id: &NodeId) {
+        if node_id.network() != self.network() {
+            tracing::warn!("Tried to ensure nostr contact for a different network {node_id}");
+            return;
+        }
+
+        if let Ok(Some(existing_contact)) = self.contact_repo.get_contact(node_id.to_owned()).await
```
codingpeanut157:
```rust
let Ok(Some(existing_contact)) = self.contact_repo.get_contact(node_id.to_owned()).await else {
continue;
};
```

crates/bcr-wallet-api/src/wallet/mod.rs
```diff
@@ -1074,24 +1086,58 @@ impl Wallet {
...
+        {
+            let existing_relays = existing_contact.nostr_relays;
+            if let Ok(fetched_relays) = self
+                .nostr_transport
+                .fetch_relay_list(node_id.npub(), existing_relays.clone())
+                .await
```
codingpeanut157, line 1116:
```rust
let Ok(fetched_relays) = self
                .nostr_transport
                .fetch_relay_list(node_id.npub(), existing_relays.clone())
                .await else { continue};

File: crates/bcr-wallet-api/src/wallet/mod.rs [hunk 4]
```diff
@@ -1074,24 +1086,58 @@ impl Wallet {
 
         Ok(res)
     }
+
+    // check relays for a given contact and update if necessary
+    async fn ensure_nostr_contact(&self, node_id: &NodeId) {
```
codingpeanut157: nitpicking

looks like a `validate` mixed with a `strengthen`, no?

codingpeanut157 (approved):

crates/bcr-wallet-api/src/wallet/mod.rs
zupzup: I mean, two wallets on different btc networks are (and should be) fundamentally incompatible.

In dev-mode, the logs are visible and once we have a notification system, we could bubble up things like that.

But for this particular case, I'm not sure, since it's just simply an invalid request, so the log is useful for debugging, but for an "actual user", I don't see how this scenario happens by accident :thinking:

PR #247, opened 2026-06-18
Small fixes and tests

zupzup:
* Tests and small fixes

### DIFF
--- a/crates/bcr-wallet-api/src/wallet/mod.rs
+++ b/crates/bcr-wallet-api/src/wallet/mod.rs
@@ -35,7 +35,7 @@ use bcr_wallet_core::{
 use bcr_wallet_persistence::{
     ContactStoreApi, NostrRepository, PaymentRequestStoreApi, TransactionRepository,
 };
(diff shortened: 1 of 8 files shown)


crates/bcr-wallet-api/src/wallet/mod.rs
```diff
@@ -90,7 +90,7 @@ impl Wallet {
         nostr_transport: Arc<dyn TransportApi>,
         nostr_event_channel: NostrEventChannel,
         nostr_repo: Arc<dyn NostrRepository>,
-        nostr_consumer: nostr::Consumer,
+        nostr_consumer: Arc<dyn ConsumerApi>,
```
codingpeanut157: is this used by multiple wallets?

codingpeanut157 (approved):

crates/bcr-wallet-api/src/wallet/mod.rs
zupzup: hm I guess not - in that case it can be a Box :+1:

PR #253, opened 2026-06-30
Wallet Names should be unique per network and dynamic dev mode

zupzup:
* Wallet Names should be unique per network
* Add endpoint to dynamically set dev-mode (without app restart)

### DIFF
--- a/crates/bcr-wallet-api/src/purse/mod.rs
+++ b/crates/bcr-wallet-api/src/purse/mod.rs
@@ -46,14 +46,17 @@ impl<Wlt> Purse<Wlt>
 where
     Wlt: WalletApi,
 {
-    pub async fn names(&self) -> Vec<String> {
+    pub async fn names_by_network(&self, network: bitcoin::Network) -> Vec<String> {
         let wallets: Vec<_> = {
             let wlts = self.wallets.read().await;
             wlts.values().cloned().collect()
         };
-        let mut res = Vec::with_capacity(wallets.len());
+        let mut res: Vec<String> = Vec::new();
         for wlt in wallets.iter() {
-            res.push(wlt.read().await.name());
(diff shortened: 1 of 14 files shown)


crates/bcr-wallet-api/src/purse/mod.rs
```diff
@@ -46,14 +46,16 @@ impl<Wlt> Purse<Wlt>
...
             let wlts = self.wallets.read().await;
             wlts.values().cloned().collect()
         };
-        let mut res = Vec::with_capacity(wallets.len());
+        let mut res: HashMap<bitcoin::Network, Vec<String>> = HashMap::new();
         for wlt in wallets.iter() {
-            res.push(wlt.read().await.name());
+            res.entry(wlt.read().await.info().network)
+                .or_default()
+                .push(wlt.read().await.name());
```
codingpeanut157: nitpicking
I have the gut feeling that doing `read` twice is actually more expensive than doing it once and query the information you need using the same lock

crates/bcr-wallet-api/src/purse/mod.rs
```diff
@@ -46,14 +46,16 @@ impl<Wlt> Purse<Wlt>
 where
     Wlt: WalletApi,
 {
-    pub async fn names(&self) -> Vec<String> {
+    pub async fn names_by_network(&self) -> HashMap<bitcoin::Network, Vec<String>> {
```
codingpeanut157: I see the only use case you have for this function is in the following call chain
```rust
purse
.names_by_network()
.await
.get(&cfg.network)
.cloned()
```
So why not simply adding a `network: bitcoin::Network` parameter, returning a `Vec<String>` and doing the filtering within the function?

codingpeanut157 (approved):

PR #256, opened 2026-07-07
Improve Client errors using new bcr-common errors

zupzup:
Improve Client errors using new bcr-common errors


crates/bcr-wallet-ffi/src/api/mod.rs
```diff
@@ -2241,7 +2265,46 @@ impl From<BcrWalletError> for WalletError {
             BcrWalletError::SwapCommitmentMismatch => {
                 WalletError::bad_request(value.to_string(), WalletErrorCode::SwapCommitmentMismatch)
             }
-            BcrWalletError::Attestation(_) => WalletError::internal(value.to_string()),
+            BcrWalletError::MintClientInternal(_) => WalletError::internal(value.to_string()),
```
stefanbitcr, line 2268: What about network errors in the mint client ? Those aren't internal wallet errors

stefanbitcr (approved):

crates/bcr-wallet-ffi/src/api/mod.rs
zupzup: I guess the question is what a network error is.

I would argue a random 500 because of a bug, or something would be an "Internal" error from the perspective of the user in that they can't do anything about that.

A 503/504 because the service is offline would/should be handled by our offline handling logic.

codingpeanut157 (approved):

PR #257, opened 2026-07-08
Don't send dleqs on same mint swaps

zupzup:
* Don't send dleq proofs on same-mint swaps to avoid giving up a blinding factor unnecessarily
* Add some tests for utils while I'm at it


crates/bcr-wallet-api/src/pocket/mod.rs
```diff
@@ -255,6 +255,8 @@ pub(crate) async fn committed_swap(
     premints: HashMap<cashu::Id, cdk00::PreMintSecrets>,
     attestation: wire_attestation::IssuanceAttestation,
 ) -> Result<Vec<cdk00::BlindSignature>> {
+    // Remove dleqs from same-mint swaps to not give up a blinding factor
+    let inputs = crate::wallet::util::remove_dleq_from_proofs(inputs);
```
stefanbitcr, line 259: ✅

stefanbitcr (approved):

PR #261, opened 2026-07-10
Fix offline substitute pay by token for attestation, multi-keysets and fees

zupzup:
* Fix offline substitute pay by token for attestation, multi-keysets and fees
* Clippy fixes for 1.97

### DIFF
--- a/crates/bcr-wallet-api/src/pocket/debit.rs
+++ b/crates/bcr-wallet-api/src/pocket/debit.rs
@@ -529,93 +529,115 @@ impl super::PocketApi for Pocket {
         &self,
         proofs: Vec<cdk00::Proof>,
         keysets_info: &HashMap<cashu::Id, KeySetInfo>,
+        keysets: HashMap<cashu::Id, KeySet>,
         client: Arc<dyn ClowderMintConnector>,
+        beta_provider: RandomBetaProvider,
         send_amount: Amount,
         swap_config: SwapConfig,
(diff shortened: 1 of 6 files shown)


crates/bcr-wallet-api/src/pocket/debit.rs
```diff
@@ -529,93 +529,115 @@ impl super::PocketApi for Pocket {
...
+        // prepare the premints
+        let mut premints: HashMap<cashu::Id, cdk00::PreMintSecrets> = HashMap::new();
+        let mut remaining_payment = send_amount;
+        // collect payments by kid, so we can reconstruct it after the swap
+        let mut payment_targets_by_kid: HashMap<cashu::Id, Amount> = HashMap::new();
+
+        for (kid, amount) in swap_plan {
+            let keyset_payment_target = std::cmp::min(amount, remaining_payment);
+
+            let target = if keyset_payment_target > Amount::ZERO {
```
codingpeanut157, line 555: If I understand this algorithm correctly, `keyset_payment_target` should never be zero, otherwise you shouldn't have included in the swap in the first place, correct?

codingpeanut157 (approved):

crates/bcr-wallet-api/src/pocket/debit.rs
zupzup: yes, at the very least it would be pointless/inefficient to include them

PR #263, opened 2026-07-13
Add DB migration, Migrate Proofs persistence to be versioned, encrypted and borsh-serialized

zupzup:
First part of the DB migration and versioning effort in Wallet-Core

* Add a Database migration scheme per wallet
* Migrate Proofs persistence to be versioned, encrypted and borsh-serialized

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 * Improve Mint Client Errors
 * Don't send dleq proofs on same-mint swaps to avoid giving up a blinding factor unnecessarily
 * Fix offline substitute pay by token for attestation, multi-keysets and fees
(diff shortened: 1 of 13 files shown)


codingpeanut157 (approved):
an intereresting benchmark on serialisation frameworks in rust

https://github.com/djkoloski/rust_serialization_benchmark

zupzup:
> an intereresting benchmark on serialisation frameworks in rust
> 
> https://github.com/djkoloski/rust_serialization_benchmark

Nice - I just included the borsh-switch here already because it seemed the most promising candidate to me, but if we should decide to use something else more broadly, now with the DB migration it's not a big issue. :+1:

PR #273, opened 2026-07-24
Counter and Commitment Migration

zupzup:
* Use latest bcr-common types and adapt to changes there
* Counter and Commitment Migration
* Next and for now final part of the migrations - the rest of the DB migrations will be done on-demand, if there's a breaking change (there's plenty of code to orient oneselves by at this point)
  * After most of the migrations are done, it might make sense to move the stuff from core/borsh.rs to bcr-common as well

### DIFF
--- a/crates/bcr-wallet-api/src/external/mint.rs
+++ b/crates/bcr-wallet-api/src/external/mint.rs
@@ -215,7 +215,7 @@ pub trait ClowderMintConnector: SendSync + std::fmt::Debug {
     async fn post_mint_onchain(
         &self,
         req: wire_mint::OnchainMintRequest,
-    ) -> Result<wire_mint::MintResponse>;
+    ) -> Result<wire_mint::OnchainMintResponse>;
     async fn post_protest_mint(
         &self,
         req: wire_mint::MintProtestRequest,
@@ -531,7 +531,7 @@ impl ClowderMintConnector for HttpClientExt {
     async fn post_mint_onchain(
         &self,
         req: wire_mint::OnchainMintRequest,
-    ) -> Result<wire_mint::MintResponse> {
+    ) -> Result<wire_mint::OnchainMintResponse> {
         let url = self
             .mint_url()
             .join(TreasuryEp::MINT_ONCHAIN_V1_EXT)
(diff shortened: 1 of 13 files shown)


crates/bcr-wallet-persistence/src/redb/pocket.rs
```diff
@@ -888,18 +910,20 @@ impl PocketRepository for PocketDB {
...
-        let old = CounterEntry { kid, counter: old };
-        let new = CounterEntry {
+        let old_c = StoredCounterPayloadV1 { kid, counter: old };
+        let old = StoredCounter::V1(old_c.clone());
+        let new = StoredCounter::V1(StoredCounterPayloadV1 {
             kid,
-            counter: old.counter + increment,
-        };
+            counter: old_c.counter + increment,
+        });
```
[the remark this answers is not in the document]
zupzup: @stefanbitcr this seems like a reasonable comment - what DO we do in that case? :thinking: 

I guess with u32::Max we have a lot of leeway, but I'll add a TODO and change this to a checked_add that fails, since the wrapping is definitely wrong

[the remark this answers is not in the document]
stefanbitcr: Good point its the wallet capacity

crates/bcr-wallet-api/src/external/mint.rs
```diff
@@ -914,7 +914,7 @@ impl ClowderMintConnector for SentinelClient {
         let res = self.secondary.post(url).json(&req).send().await?;
         match res.error_for_status_ref() {
             Ok(_) => {
-                let response: wire_mint::MintResponse = res.json().await?;
+                let response: wire_mint::EbillMintResponse = res.json().await?;
```
stefanbitcr: I don't think these are correct or that EbillMintResponse is intended to be used here,

We might need a separate Response here for onchain even if the ebill type fits ?

Keep for now but note we probably change this as its strange to use Ebill Mint Response for an Onchain Mint. We have it wrongly in Clowder as well

zupzup: Right, I was just fixing the error caused by the rename - but I guess there should be an `OnchainMintResponse` as well.
Should I add one to bcr-common?

stefanbitcr: Yeah if you are busy with it

stefanbitcr (approved):

PR #274, opened 2026-07-27
Transactions Rework & Migration & Reclaim Handling

zupzup:
* Rework Transactions Data Model
* Implement Data Migration to new Transactions Data Model
* Fix some bugs with payment requests and relays
* Rework reclaim to create two linked transactions - one `Settled` and one `Canceled`

### DIFF
--- a/crates/bcr-wallet-persistence/src/lib.rs
+++ b/crates/bcr-wallet-persistence/src/lib.rs
@@ -7,11 +7,13 @@ pub mod test_utils;
 use crate::error::Result;
 use async_trait::async_trait;
 use bcr_common::cashu::{self, nut00 as cdk00, nut01 as cdk01, nut07 as cdk07};
-use bcr_common::cdk_common::wallet::{Transaction, TransactionId};
 use bcr_common::core::NodeId;
 use bcr_wallet_core::contact::Contact;
 use bcr_wallet_core::name::Name;
-use bcr_wallet_core::types::{PaymentRequestDirection, PaymentRequestState};
+use bcr_wallet_core::types::{
+    PaymentRequestDirection, PaymentRequestState, Transaction, TransactionLinkReason,
+    TransactionStatus,
+};
 use bcr_wallet_core::{
     SendSync,
     types::{PaymentRequest, WalletConfig},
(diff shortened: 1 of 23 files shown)


crates/bcr-wallet-persistence/src/redb/payment_request.rs
```diff
@@ -41,7 +40,7 @@ impl From<PaymentRequestEntryDirection> for PaymentRequestDirection {
 #[derive(Debug, Clone, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
 pub enum PaymentRequestEntryState {
     Pending,
-    Paid { tx_id: TransactionId },
+    Paid { tx_id: Uuid },
```
stefanbitcr, line 43: (minor) Breaking change ?

crates/bcr-wallet-persistence/src/redb/transaction.rs
```diff
@@ -126,34 +252,46 @@ impl TransactionDB {
...
-    ) -> Result<()> {
-        let write_txn = db.begin_write()?;
+        tx_id_to_find: TransactionId,
+    ) -> Result<Option<Transaction>> {
+        let read_txn = db.begin_read()?;
 
-        {
-            let mut table = write_txn.open_table(tx_table)?;
-            table.remove(tx_id.as_bytes().as_slice())?;
+        match read_txn.open_table(tx_table) {
```
stefanbitcr: This is a full scan for every receive

crates/bcr-wallet-persistence/src/lib.rs
```diff
@@ -95,24 +97,24 @@ pub trait PurseRepository: SendSync {
...
-        tx_id: TransactionId,
-        key: String,
-        value: String,
-    ) -> Result<Option<String>>;
-    async fn update_memo(
+        tx_id: Uuid,
+        status: TransactionStatus,
+    ) -> Result<Option<TransactionStatus>>;
+    async fn update_memo(&self, tx_id: Uuid, new_memo: Option<String>) -> Result<Option<String>>;
+    async fn update_fee(&self, tx_id: Uuid, fee_to_add: cashu::Amount) -> Result<()>;
```
stefanbitcr: Is `update_fee` dead code ?

stefanbitcr (approved):

crates/bcr-wallet-persistence/src/redb/payment_request.rs
zupzup: Yes, but this hasn't been implemented so far on frontend in any version, so it should be safe.

crates/bcr-wallet-persistence/src/lib.rs
zupzup: yes, will remove :+1:

crates/bcr-wallet-persistence/src/redb/transaction.rs
zupzup: Yeah I know, we'd need an index from cdk_common::wallet::TransactionId -> Uuid (transaction ID) to make this basically free, this is one of many perf improvements we need to do at some time, but should be fine for the time being.

However, I will check if we even need this code path anymore, because this was kind of a workaround for not being able to have 2 transactions for this (same y's etc), which is not true anymore, so maybe this case is just an incoming tx now and not a reclaim, then I could just get rid of it

zupzup: Removed it :+1:

Issue #228, opened 2026-06-01
Recovery and multic-device sync via Nostr


mtbitcr:

zupzup:
TODO: research other existing wallets

zupzup:
If we add nostr support, we can theoretically store all our tokens on the nostr network. Thus by npub synchronise all our devices.

zupzup:
We must be able to restore wallet from local file and from nostr.

zupzup:
Can we somehow sync the counter for proofs? what if e.g. 2 wallets use the same seed phrase but on different mints?

PR #277, opened 2026-07-30
Rework Contacts and Payment Requests

zupzup:
* Rework Contacts
    * Fields
        * id is now a `Uuid`
        * node_id is an `Option<NodeId>`
        * add email as an `Option<Email>`
        * name is an `Option<Name>`
        * add company as an `Option<Name>`
        * Either company, or name have to be set
        * Either node_id, or email have to be set
        * For `edit_contact`
            * All fields have to be provided
            * To delete a field, send it as `None`
            * To set a field, send it as `Some(new_value)`
    * Contacts are now on `purse` level instead of on `wallet` level and are persisted per `btc_network`
        * there is a contact book for each btc network
    * DB migration & versioning for contacts
    * It would be good to add all clowder-relays (wildcat0-n) to the wallet config
* Payment Request functions and pay by contact take a `Uuid` (the contact id) now
    * Contacts have to have a `node_id` to be eligible for `pay_to_contact` or `request_payment_from_contact`
* Expose functionality to send a private, encrypted payment to an existing contact via Nostr
    * `wallet_prepare_pay_to_contact` - taking a contact_id, amount and description
    * `wallet_pay_to_contact` - taking a payment request id

### DIFF
--- a/crates/bcr-wallet-api/src/wallet/api.rs
+++ b/crates/bcr-wallet-api/src/wallet/api.rs
@@ -21,9 +21,7 @@ use bcr_common::{
 };
 use bcr_wallet_core::{
     SendSync,
-    contact::Contact,
     event::{ContactPaymentRequestPayload, EventEnvelope},
-    name::Name,
     types::{
         MeltEstimation, PaymentRequest, PaymentRequestDirection, PaymentRequestState,
         PaymentResultCallback, PaymentType, PendingPaymentSubscriptionCallback, Transaction,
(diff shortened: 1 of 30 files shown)


crates/bcr-wallet-api/src/wallet/api.rs
stefanbitcr, line 1277: Does this delete contacts network wide for the wallet ?

stefanbitcr (approved):
How are the old contacts migrated ?

crates/bcr-wallet-api/src/wallet/api.rs
zupzup: No, not for now - there's a future ticket for multi-device sync [here](https://github.com/BitcreditProtocol/Wallet-Core/issues/228) but it's unclear if this will also migrate contacts (in E-Bill we don't currently)

zupzup:
> How are the old contacts migrated ?

They aren't - there are no released versions with contacts (only some local dev ones), so no one has data to migrate.

PR #278, opened 2026-07-30
Separate Fees into multiple Fee Types

zupzup:
* Separate `fees` in `Transaction` and `PaymentSummary` to contain different types of fees
    * `swap` - swap fee
    * `network` - network fee (e.g. bitcoin miner fee)
    * `melt` - melt fee

### DIFF
--- a/crates/bcr-wallet-api/src/pocket/debit.rs
+++ b/crates/bcr-wallet-api/src/pocket/debit.rs
@@ -821,7 +821,7 @@ impl DebitPocketApi for Pocket {
 
         // inputs need to cover amount + network_fee + melt_fee
         let full_amount = amount + network_fee + melt_fee;
-        let (_, send_ref) = self
+        let (send_summary, send_ref) = self
             .compute_send_costs(Amount::from(full_amount), keysets_info)
             .await?;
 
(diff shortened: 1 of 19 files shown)


crates/bcr-wallet-api/src/pocket/debit.rs
stefanbitcr, line 825: Unrelated, but does this skip swap fees when when selected proofs don't match exactly ?

stefanbitcr (approved):

crates/bcr-wallet-api/src/pocket/debit.rs
zupzup: Yeah that's a good point - I'll check it out :+1:

crates/bcr-wallet-persistence/src/redb/migration/migration_0003.rs
```diff
@@ -101,18 +101,39 @@ impl TryFrom<TransactionEntry> for Transaction {
     type Error = Error;
 
     fn try_from(entry: TransactionEntry) -> Result<Self> {
+        let pt = get_payment_type(&entry.metadata);
```
stefanbitcr, line 104: Do the split fees get flattened again ?

zupzup: No, this is just an update to the previous migration from the old TransactionEntry to the V1, which both have a flat fee, but so a migration from a very old app to the newest V2 works as well (but without duplicating the `Transaction` struct in code)

crates/bcr-wallet-api/src/pocket/debit.rs
zupzup: Yeah, good catch - the accounting for melt was actually wrong. The amount included the fees and the swap fee was not accounted for (the balance updated correctly, of course, but the transaction had the wrong values) - will fix :+1:

PR #287, opened 2026-08-03
Add endpoints for sharing payment requests

zupzup:
* Add endpoints to create a shareable payment request and to pay such a request
    * `wallet_create_shareable_remote_payment_request`
        * takes `wallet_id`, `amount`, `description`
        * returns a string (base58 encoded, borsh-serialized payment request)
    * `wallet_prepare_pay_shared_payment_request`
        * takes `wallet_id`, `payment_request` (the string from above)
        * parses the request and prepares the given payment
        * returns `payment_summary`
    * `wallet_pay_shared_payment_request`
        * takes `wallet_id`, `payment_request_id` (the string from above)
        * executes the payment
    * In the case of a shareable payment request, no payment request is actually persisted in the wallet

### DIFF
--- a/crates/bcr-wallet-api/src/wallet/api.rs
+++ b/crates/bcr-wallet-api/src/wallet/api.rs
@@ -143,7 +143,16 @@ pub trait WalletApi: SendSync {
         relays: Vec<RelayUrl>,
     ) -> Result<Vec<RelayUrl>>;
     async fn delete(&self) -> Result<()>;
-    // pending payment requests
+    async fn create_shareable_remote_payment_request(
+        &self,
+        amount: Amount,
+        unit: CurrencyUnit,
+        description: Option<String>,
+    ) -> Result<String>;
+    async fn prepare_pay_shared_payment_request(
+        &self,
+        payment_req: String,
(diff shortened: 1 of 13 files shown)


crates/bcr-wallet-api/src/wallet/api.rs
```diff
@@ -1278,6 +1333,73 @@ impl WalletApi for super::Wallet {
...
+            None,
+            to_mint_url(self.client.mint_url()),
+        );
+        let event: EventEnvelope =
+            bcr_wallet_core::event::Event::new_contact_payment_request(payload).try_into()?;
+        let encoded_payload = base58::encode(&borsh::to_vec(&event)?);
+        Ok(encoded_payload)
+    }
+
+    async fn prepare_pay_shared_payment_request(
```
stefanbitcr, line 1356: Do we need to check the network here ?

stefanbitcr (approved):

crates/bcr-wallet-api/src/wallet/api.rs
zupzup: yes, good point :+1:

PR #289, opened 2026-08-04
Temporary Foreign-Mint-Funds Storage

zupzup:
* Up to 0.9.10
* Store change from offline-tokens into a temporary-foreign-mint-proof-storage
  * these foreign-mint-proofs are regularly attempted to reclaim in a job

### DIFF
--- a/crates/bcr-wallet-api/src/wallet/mod.rs
+++ b/crates/bcr-wallet-api/src/wallet/mod.rs
@@ -25,9 +25,9 @@ use bcr_wallet_core::{
     contact::Contact,
     event::{ContactPaymentPayload, EventEnvelope},
     types::{
-        ListTransactionsResult, PaymentRequest, PaymentType, Transaction, TransactionCursor,
-        TransactionFees, TransactionFilters, TransactionLinkReason, TransactionSort,
-        TransactionStatus, extract_fees_per_month,
+        ClowderBeta, ForeignMintProof, ListTransactionsResult, PaymentRequest, PaymentType,
+        Transaction, TransactionCursor, TransactionFees, TransactionFilters, TransactionLinkReason,
+        TransactionSort, TransactionStatus, extract_fees_per_month,
     },
     util::{from_mint_url, to_mint_url},
 };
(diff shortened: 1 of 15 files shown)


crates/bcr-wallet-persistence/src/redb/pocket.rs
```diff
@@ -686,6 +808,72 @@ impl PocketDB {
...
+                let key = foreign_mint_proof_key(&clowder_id, y);
+                table.remove(key.as_slice())?;
+            }
+        }
+
+        write_txn.commit()?;
+        Ok(())
+    }
+
     fn delete_repo(
```
stefanbitcr, line 877: should we drop the table so the secrets are removed

crates/bcr-wallet-api/src/wallet/mod.rs
```diff
@@ -590,6 +590,78 @@ impl Wallet {
...
+                    map.entry(item.clowder_id).or_default().push(item);
+                    map
+                });
+
+        let mut reclaimed = Amount::ZERO;
+
+        // create tokens and attempt to reclaim
+        // if it's from our own mint, create a token and swap it to be safe
+        // if it's from another mint, create a token and do an intermint swap
+        for mint in all_mints.iter() {
```
stefanbitcr, line 622: Does a problematic proof make the batch fail indefinitely ?

stefanbitcr (approved):

crates/bcr-wallet-persistence/src/redb/pocket.rs
zupzup: oh yes definitely, good catch :+1:

crates/bcr-wallet-api/src/wallet/mod.rs
zupzup: Yeah, I considered that as well, but since this is more of a temporary fix for the demo, I didn't go all the way.

I thought an improvement would be to check the status of all proofs before adding it to the token and only use the unspent ones. :thinking: I'll create an improvement task in case we keep this solution for longer

It's unclear what we would do with proofs we can't reclaim - I guess we'd log it and once we have notifications somehow tell the user?

zupzup: Task: https://github.com/BitcreditProtocol/Wallet-Core/issues/293

